PR #92, opened 2023-06-23
Fix some app startup issues 

simonwoerpel:
:wave: 

this PR includes 2 small bugfixes:
1.) startup app even if no `NOTIFY_SMTP_PORT` is given as env var
2.) fix the `pgp_owner_info` endpoint for pgp keys that have no expiration date


glenn-sorrentino:
Hey, thanks, Simon!

PR #138, opened 2023-10-11
remove messages.txt file

sts10:
Guessing this `./messages.txt` file was a hold-over from an older system for Hush Line where messages would be decrypted to a text file?

### DIFF
--- a/messages.txt
+++ b/messages.txt
@@ -1,2 +1,0 @@
-👇 Encrypted messages will appear here after they're submitted.
-


glenn-sorrentino:
This is the file where encrypted messages are saved upon submission.

sts10:
Oh wow, I didn't know about that. Very interesting... especially if we want to build an email-free version of Hush Line.

PR #144, opened 2023-10-13
`reset` terminal window before first whiptail prompt

sts10:
This change fixes a bug I was experiencing on my Pi (Bookworm) in which the Hush Line install script would get "stuck" on the first whiptail prompt (I was unable to get past it by hitting enter). 

Running the simple `reset` command, with its default flags, before the first whiptail call appears to fix the issue for me. 

I don't think the command will have any adverse side effects for users who were not having my issue with the whiptail prompt(s).

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -44,6 +44,9 @@ else
     cd $HOME # return to HOME for next steps
 fi
 
+# "reset" the terminal window before running first whiptail prompt
+reset
+
 # Welcome Prompt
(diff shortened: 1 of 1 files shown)


sts10:
See `man reset` for some more info: 
```
   reset - reinitialization
       When invoked as reset, tset sets the terminal modes to “sane” values:

       •   sets cooked and echo modes,

       •   turns off cbreak and raw modes,

       •   turns on newline translation and

       •   resets any unset special characters to their default values

       before doing the terminal initialization described above.  Also, rather
       than  using  the  terminal initialization strings, it uses the terminal
       reset strings.

       The reset command is useful after a program dies leaving a terminal  in
       an abnormal state:

       •   you may have to type

               <LF>reset<LF>

           (the line-feed character is normally control-J) to get the terminal
           to work, as carriage-return may no  longer  work  in  the  abnormal
           state.

       •   Also, the terminal will often not echo the command.
```

glenn-sorrentino (approved):

PR #146, opened 2023-10-13
POC: Auto-complete gmail smtp info in web form

sts10:
My Javascript skills are a bit rusty (and **I doubt I put this javascript where a Flask app wants it**) but I thought, as a proof-of-concept, I'd attempt to have the Hush LineBlackbox web form auto-complete the smtp info for a gmail address. 

It does this by listening for a change in the email field and, if the value includes "@gmail.com", it writes the default gmail server URL and port number into the relevant fields.

Just an idea!

### DIFF
--- a/templates/setup.html
+++ b/templates/setup.html
@@ -46,5 +46,17 @@
         
         <button type="submit">Submit</button>
     </form>
+    <script>
+        // automatically fill in gmail smtp info if user enters @gmail.com into the email field
+        document.getElementsByName("email")[0].addEventListener("change", (event) => {
+          if (event.isComposing || event.keyCode === 229) {
+            return;
+          }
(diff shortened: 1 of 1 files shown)


glenn-sorrentino:
Did you mean to put this in the Blackbox repo?

sts10:
> Did you mean to put this in the Blackbox repo?

I definitely would have thought that's where this particular web form HTML template would live, but it seems to be in the `hushline` repo? And I don't see any HTML templates in the blackbox repo. Is that a mistake of file organization?

glenn-sorrentino:
i actually just reorganized. i moved setup.html to blackbox since it's specific to that project

sts10:
See https://github.com/scidsg/blackbox/pull/35

PR #184, opened 2023-11-08
Modify Close App button link to main Wikipedia page

sN4Kec4Se:
Proposition to change the link redirection on the "Close App" button from a random Wikipedia topic page to the main Wikipedia start page: https://www.wikipedia.org 

This would reduce the risk of an "unlucky roll" where a user might get a random Wikipedia page topic that in context might be inappropriate or even dangerous to the situation. 

It is a relatively low risk and the randomness was a creative feature but it would likely be safer to keep the redirection result constant.


glenn-sorrentino:
Thank you, @sN4Kec4Se!

PR #237, opened 2024-02-26
Update install.sh

ritik2311:
1. Poetry Installation: Installs the poetry command-line tool.
2. Poetry Shell: Creates a dedicated project environment using poetry shell.
3. Dependency Management: Installs project dependencies using poetry install commands.


ritik2311:
Updated the install.sh with poerty so no need for this now

Issue #280, opened 2024-03-19
App import should not list GPG keys


brassy-endomorph:
It is unclear why this is happening despite the code comment, and we likely do not want to list keys like this. If we need to, we can hide it behind a `Click` command optionally gated behind a debug env var so like `HUSHLINE_DEBUG_OPTS=1 flask debug list-keys` or something.

https://github.com/scidsg/hushline/blob/75ee6e0e58e71f014aedf4381dbd377d38691de7/app.py#L778-L779

glenn-sorrentino:
Closed by #292

PR #292, opened 2024-03-20
Organize Debug Commands into CLI Command Group

glenn-sorrentino:
## Changes

- Encapsulated the `list_keys` function within a Flask CLI command group named `debug`. This helps organize debugging-related commands and keeps the main command namespace clean.

## Why

- Improves discoverability and organization of debugging commands without cluttering the primary command space.

### DIFF
--- a/hushline/__init__.py
+++ b/hushline/__init__.py
@@ -56,10 +56,11 @@ def create_app() -> Flask:
     for module in [admin, settings]:
         app.register_blueprint(module.create_blueprint())
 
-    # Correctly initialize the debug_cli AppGroup before defining commands
-    debug_cli = AppGroup("debug", help="Debugging related commands")
+    @app.cli.group(help="Debugging related commands")
+    def debug() -> None:
+        pass
 
-    @debug_cli.command("list-gpg-keys")
+    @debug.command("list-gpg-keys")
     def debug_list_gpg_keys():
         """List GPG keys for debugging."""
         if os.getenv("HUSHLINE_DEBUG_OPTS") == "1":
@@ -67,7 +68,7 @@ def create_app() -> Flask:
         else:
             print("Debugging options are not enabled. Set HUSHLINE_DEBUG_OPTS=1 to enable.")
 
-    app.cli.add_command(debug_cli)
+    app.cli.add_command(debug)
 
     @app.errorhandler(404)
     def page_not_found(e):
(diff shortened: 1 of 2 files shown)


hushline/__init__.py
```diff
@@ -54,6 +56,19 @@ def handle_rate_limit_exceeded(e):
     for module in [admin, settings]:
         app.register_blueprint(module.create_blueprint())
 
+    # Correctly initialize the debug_cli AppGroup before defining commands
+    debug_cli = AppGroup("debug", help="Debugging related commands")
```
brassy-endomorph, line 60: This line plus the `app.cli.add_command(debug_cli)` can just be done like this:

```python
@app.cli.group(help="Debugging related commands")
def debug() -> None:
    pass
```

This is how the docs recommend it, and it seems to be preferrable, but this is just a nit.

brassy-endomorph (approved):

hushline/__init__.py
glenn-sorrentino: New update pushed

brassy-endomorph:
Ok, one last comment that I'm gonna refer to as "`git` discipline" but we should avoid merge commits from the base branch and should instead rebase. And to the extend possible we should interactively rebase to squash fixups into their relevant commits. This makes it so much easier to track where a bug came from, and having a clean git history means that all the tooling for showing history and branches actually represents things a bit better. This is an easy change to make for this PR. I think the steps are:

- `git reset HEAD^ --hard` to remove the merge
- `git rebase -i 6d6c1f5^` and then make `b8bcf83` a fixup on top of `6d6c1f5`
- Rebase your branch on main if there's merge conficts, but there shouldn't be?
- Force push

glenn-sorrentino:
> Ok, one last comment that I'm gonna refer to as "`git` discipline" but we should avoid merge commits from the base branch and should instead rebase. And to the extend possible we should interactively rebase to squash fixups into their relevant commits. This makes it so much easier to track where a bug came from, and having a clean git history means that all the tooling for showing history and branches actually represents things a bit better. This is an easy change to make for this PR. I think the steps are:
> 
>     * `git reset HEAD^ --hard` to remove the merge
> 
>     * `git rebase -i 6d6c1f5^` and then make `b8bcf83` a fixup on top of `6d6c1f5`
> 
>     * Rebase your branch on main if there's merge conficts, but there shouldn't be?
> 
>     * Force push

Done. Thanks for the guidance!

glenn-sorrentino:
Closes #280

Issue #279, opened 2024-03-19
User and SecondaryUser are ambiguous


brassy-endomorph:
Without docs, it's not immediately clear what a `User` vs. a `SecondaryUser` are. It _seems_ like one is `Admin` and the other is `Source` (or `Leaker` or whatever). The names should more accurately reflect what they are and do.

glenn-sorrentino:
Yeah, the `SecondaryUser` table is actually for secondary usernames. 

It's a paid feature tailored for organizational use cases with the ability to add up to 5 additional usernames so they can all be managed from the same account.

PR #295, opened 2024-03-22
Renamed SecondaryUser to SecondaryUsername

glenn-sorrentino:
## What's Changed
Renamed the `SecondaryUser` model to `SecondaryUsername`. This is to make it clearer what the model is about. It's not about users; it's about their usernames.

## Specific Changes
- Renamed the class from `SecondaryUser` to `SecondaryUsername`.
- Updated the table name in the DB from `secondary_user` to `secondary_username`.
- Fixed all references to this model in the code, including relationships and foreign keys. Wherever we had `secondary_users`, it's now `secondary_usernames`.
- Updated foreign keys in the `Message` model to link to `secondary_username.id` instead of `secondary_user.id`.
- Adjusted the backref in the `User` model to match the new naming.

## Why
The old name was misleading. This change makes our model names match what they actually represent.

## Next Steps
- Test the parts of the app that use this model to make sure nothing broke.

### DIFF
--- a/admin.py
+++ b/admin.py
@@ -2,7 +2,7 @@
 
 import sys
 
-from hushline import SecondaryUser, User, app, db
+from hushline import SecondaryUsername, User, app, db
 
 
 def toggle_admin(username):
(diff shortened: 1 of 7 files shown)


glenn-sorrentino:
Closes #279

brassy-endomorph (approved):

Issue #275, opened 2024-03-19
Use `passlib` to hash/verify passwords


brassy-endomorph:
We should be using [`passlib`](https://passlib.readthedocs.io/en/stable/) for hashing passwords and verifying hashes. This gives us a pluggable interface for eventually upgrading hashes.

PR #296, opened 2024-03-23
Refactor User model to use Passlib for secure password hashing

glenn-sorrentino:
- Introduced Passlib's CryptContext for hashing passwords using bcrypt scheme. Ensured backward compatibility by marking bcrypt as the default scheme and enabling auto deprecation for future upgrades.
- Modified the `password_hash` property setter in the User model to hash plaintext passwords with Passlib before encrypting and storing the hash. This ensures that passwords are securely managed and stored in an encrypted and hashed format, enhancing security.
- Added a `verify_password` method to the `User` model for decrypting and verifying password hashes against plaintext passwords using Passlib. This method supports secure password verification during user authentication processes.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,12 +16,12 @@ flask-migrate = "^4.0.7"
 flask-sqlalchemy = "^3.1.1"
 flask-wtf = "^1.2.1"
 gnupg = "^2.3.1"
+passlib = "^1.7.4"
 pymysql = "^1.1.0"
 pyotp = "^2.9.0"
 python-dotenv = "^1.0.1"
 qrcode = "^7.4.2"
 redis = "^5.0.3"
-passlib = "^1.7.4"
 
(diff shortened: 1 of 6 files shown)


glenn-sorrentino:
Closes #275

pyproject.toml
```diff
@@ -21,6 +21,7 @@ pyotp = "^2.9.0"
 python-dotenv = "^1.0.1"
 qrcode = "^7.4.2"
 redis = "^5.0.3"
+passlib = "^1.7.4"
```
brassy-endomorph, line 24: Nit: it would be good to keep the dependencies alphabetized

brassy-endomorph (approved):

pyproject.toml
glenn-sorrentino: Updated 👍

PR #299, opened 2024-03-25
Remove MariaDB

glenn-sorrentino:
- Updated `SQLALCHEMY_DATABASE_URI` in the application configuration to use SQLite3.
- Removed `pymysql` dependency from `pyproject.toml` as it's no longer needed.


hushline/__init__.py
```diff
@@ -28,7 +28,7 @@ def handle_rate_limit_exceeded(e):
 
     app.config["SECRET_KEY"] = os.getenv("SECRET_KEY")
 
-    app.config["SQLALCHEMY_DATABASE_URI"] = os.environ["SQLALCHEMY_DATABASE_URI"]
+    app.config["SQLALCHEMY_DATABASE_URI"] = "sqlite:///hushline.db"
```
brassy-endomorph, line 31: This should still be set by the env var no matter what DB we use.

pyproject.toml
```diff
@@ -17,7 +17,6 @@ flask-sqlalchemy = "^3.1.1"
 flask-wtf = "^1.2.1"
 gnupg = "^2.3.1"
 passlib = "^1.7.4"
-pymysql = "^1.1.0"
```
brassy-endomorph, line 20: The lockfile needs to be updated too as I'm fairly sure doing so will yank the `pymysql` deps and a few others

brassy-endomorph (changes requested):
I don't think we need this PR yet, and I think it would be better to wait until we know what sort of database we want and why before we commit to more changes.

brassy-endomorph (dismissed):

hushline/__init__.py
glenn-sorrentino: Updated installer to remove MariaDB and include sqlite in the .env config

install.sh
```diff
@@ -190,8 +190,7 @@ fi
 
 if ! egrep -q '^SQLALCHEMY_DATABASE_URI=' .env; then
     echo 'Setting SQLALCHEMY_DATABASE_URI'
-    # It's assumed DB_NAME, DB_USER, DB_PASS have been already captured above
-    echo "SQLALCHEMY_DATABASE_URI=mysql+pymysql://$DB_USER:$DB_PASS@localhost/$DB_NAME" >> .env
+    echo "SQLALCHEMY_DATABASE_URI=sqlite:///hushline.db" >> .env
```
brassy-endomorph, line 193: This should really have an absolute path like `/var/lib/hushline/sqlite.db` so it's not sitting in the app's code directory

brassy-endomorph (commented):
need to remove the python mysql dependency and recreate the lockfile

brassy-endomorph:
Also can you set up your IDE to include trailing newlines on all the files to reduce noise in the commits?

install.sh
glenn-sorrentino: Added absolute path:
`echo "SQLALCHEMY_DATABASE_URI=sqlite:////var/lib/hushline/hushline.db" >> .env`

Added DB folder path and permissions:
```
# SQLite database directory and file path
SQLITE_DIR="/var/lib/hushline"
SQLITE_DB_FILE="$SQLITE_DIR/hushline.db"

# Ensure SQLite directory exists
if [ ! -d "$SQLITE_DIR" ]; then
    echo "Creating SQLite directory at $SQLITE_DIR..."
    mkdir -p "$SQLITE_DIR"
    # Ensure the directory is owned by the application user
    chown $HUSHLINE_USER:$HUSHLINE_GROUP "$SQLITE_DIR"
    echo "✅ SQLite directory created."
else
    echo "👍 SQLite directory already exists."
fi
```

brassy-endomorph:
Is it possible to rebase the commits to squash them down to discrete units of work? There's a lot of seeming duplicates. It looks like you removed MariaDB 3 times according the commit

glenn-sorrentino:
> Is it possible to rebase the commits to squash them down to discrete units of work? There's a lot of seeming duplicates. It looks like you removed MariaDB 3 times according the commit

Yeah, I think it got messy when I rebased it, as this was an older branch.

I squashed a handful into two commits. Trying on others gave:

> Unable to squash. Squashing replays all commits up to the last one required for the squash. A merge commit cannot exist among those commits.

However, a merge wasn't in the group selected.

brassy-endomorph (approved):

PR #306, opened 2024-03-28
added mypy

### LINKED ISSUE #264: Add mypy type checker
We should add a non-strict set of rules from `mypy` to add typechecking to the app to prevent common errors

brassy-endomorph:
- Added `mypy` to `Makefile` 
- Added annotations to the files
- Left things fairly non-strict so that annoying errors simply aren't dealt with yet

Fixes #264


glenn-sorrentino:
I tried to retrofit these changes to the Sequoia branch but was unsuccessful. That branch is now merged if you want to use the latest.

glenn-sorrentino (changes requested):
Apologies if this is a time kill, but can you try these updates with the newest Main branch?

Issue #283, opened 2024-03-19
The python package gnupg is unmaintained


brassy-endomorph:
This needs to be replaced, possibly with [`pysequoia`](https://pypi.org/project/pysequoia/)

glenn-sorrentino:
Closed after adding pysequoia

PR #308, opened 2024-03-28
Add pyseqouia

glenn-sorrentino:
- Successfully added `pyseqouia`
- Include `rust` in installer
- Addressed an issue where client-side encryption was only happening when logged in
- Removed `gnupg`
- Removed `list_keys`


glenn-sorrentino:
This closes #283 

Merging as this is verified to be working, and the number of changes here impacts your PR, @brassy-endomorph. I tried retrofitting with your changes but with no success.

PR #318, opened 2024-04-03
Integrate Ansible

glenn-sorrentino:
- Added setup.yml
- Added common.yml
- Added encrypted secrets/secrets.yml
- Installed Ansible locally and tested deploying.
   - `ansible-playbook -i <IP>, setup.yml --ask-vault-pass -e ansible_ssh_private_key_file=~/.ssh/dev -e ansible_user=root`
    - [Gist of last install](https://gist.github.com/glenn-sorrentino/4f2cc83825d06c63afdf79d43d0dc5e8). Hitting a persistent issue where installing dependencies with Poetry fails. 
       - [When manually attempting installation](https://gist.github.com/glenn-sorrentino/4b656e67ea664128b3f53d2ea084c129), I activated the environment and executed the installation command in the playbook and it succeeds.


glenn-sorrentino:
> * Added setup.yml
> 
>     * Added common.yml
> 
>     * Added encrypted secrets/secrets.yml
> 
>     * Installed Ansible locally and tested deploying.
>       
>       * `ansible-playbook -i <IP>, setup.yml --ask-vault-pass -e ansible_ssh_private_key_file=~/.ssh/dev -e ansible_user=root`
>       * [Gist of last install](https://gist.github.com/glenn-sorrentino/4f2cc83825d06c63afdf79d43d0dc5e8). Hitting a persistent issue where installing dependencies with Poetry fails.
>         
>         * [When manually attempting installation](https://gist.github.com/glenn-sorrentino/4b656e67ea664128b3f53d2ea084c129), I activated the environment and executed the installation command in the playbook and it succeeds.

Resolved by activating the venv, then calling poetry by it full path:

```
- name: Install dependencies with Poetry
      shell: |
        cd /var/www/html/hushline && source /var/www/html/hushline/venv/bin/activate && /root/.local/bin/poetry install
      args:
        executable: /bin/bash
```

install.sh
brassy-endomorph, line 1: Ansible should fully replace `install.sh`

migrations/versions/577b927c0f49_.py
brassy-endomorph, line 1: Migrations shouldn't be part of this PR.

pyproject.toml
```diff
@@ -47,6 +48,7 @@ line_length = 100
 pythonpath = "."
 
 [tool.mypy]
+python_version = "3.11"
```
brassy-endomorph, line 51: Why was this added?

setup.yml
```diff
@@ -0,0 +1,174 @@
...
+          - redis
+          - redis-server
+          - apt-transport-https
+          - curl
+          - sqlite3
+          - acl
+        update_cache: yes
+        state: latest
+
+    - name: Clone Hush Line repository
```
brassy-endomorph, line 32: This shouldn't be a clone. That implies that we're taking the latest `HEAD` version of the default branch (currently `main`). All the files are local to the repo and should be copied over with `rsync` or something similar.

setup.yml
```diff
@@ -0,0 +1,174 @@
...
+      git:
+        repo: https://github.com/scidsg/hushline.git
+        dest: /var/www/html/hushline
+        version: ansible
+
+    - name: Install Poetry
```
brassy-endomorph, line 38: This should be installed via `apt`

setup.yml
```diff
@@ -0,0 +1,174 @@
...
+    - name: Create virtual environment
+      shell: python3 -m venv /var/www/html/hushline/venv
+      args:
+        creates: /var/www/html/hushline/venv
+
+    - name: Configure Poetry to use existing virtual environment
+      shell: |
+        /root/.local/bin/poetry config virtualenvs.create false
+      args:
+        executable: /bin/bash
```
brassy-endomorph, line 53: We don't need to specify `bash` each time.

setup.yml
```diff
@@ -0,0 +1,174 @@
...
+
+    - name: Install dependencies with Poetry
+      shell: |
+        source /var/www/html/hushline/venv/bin/activate && /root/.local/bin/poetry install
+      args:
+        chdir: /var/www/html/hushline
+        executable: /bin/bash
+
+    - name: Perform database migrations
```
brassy-endomorph, line 62: This shouldn't be done in ansible but rather `systemd` via `ExecStartPre`

setup.yml
```diff
@@ -0,0 +1,174 @@
...
+      shell: |
+        source /var/www/html/hushline/venv/bin/activate && flask db upgrade
+      args:
+        chdir: /var/www/html/hushline
+        executable: /bin/bash
+      become: no
+      become_user: hushlineuser
```
brassy-endomorph, line 69: `become_user` used before user creation

setup.yml
```diff
@@ -0,0 +1,174 @@
...
+        owner: hushlineuser
+        group: www-data
+
+    - name: Ensure /etc/hushline directory exists
+      file:
+        path: /etc/hushline
+        state: directory
+        owner: hushlineuser
+        group: www-data
+        mode: '0755'
```
brassy-endomorph, line 92: Mode `755` is unsafe as this directory will have sensitive configurations. Should be `0700`

vars/secrets/secrets.yml
brassy-endomorph, line 1: Secret values for our private instance shouldn't be in the FOSS repo

vars/common.yml
```diff
@@ -0,0 +1,12 @@
+hushline_user: hushlineuser
```
brassy-endomorph, line 1: We have a var `hushline_user` set here but we hardcode the string `hushlineuser` in a bunch of places.

1. We should check that vars are reused and not hardcoded
2. The user should just be `hushline` since the `user` part is implied

hushline.nginx.j2
```diff
@@ -0,0 +1,46 @@
...
+    include snippets/ssl-params.conf;
+
+    # Root directory
+    root /var/www/html/hushline;
+
+    # Proxy pass configuration for handling requests
+    location / {
+        proxy_pass http://unix:/var/www/html/hushline/hushline.sock;
+        proxy_set_header Host $host;
+        proxy_set_header X-Real-IP $remote_addr;
```
brassy-endomorph, line 26: If we're using CF for production, we don't want to `set` these headers since the real IP would be set to the CF IPs and thus would break all our rate limiting

brassy-endomorph (changes requested):
a few things to start, and a few more comments in the other PR that this should be rebased on top of once that's merged

glenn-sorrentino:
Closing this PR as @brassy-endomorph suggested taking on this work.

setup.yml
glenn-sorrentino: I don't think poetry is available with `apt`. Their documentation suggests that the curl command is the "official" method: https://python-poetry.org/docs/#installing-with-the-official-installer

hushline.nginx.j2
glenn-sorrentino: Resolving as we've moved, at least for now, away from CF.

vars/secrets/secrets.yml
glenn-sorrentino: added to gitignore

PR #321, opened 2024-04-07
Dev Workflow Using `Makefile`

glenn-sorrentino:
- Moved environment variables to `env.sh`
- Updated dev documentation
- Remove broken package references
- Updated password hashing bug blocking registration


ritik2311 (approved):
Locally tested and run without any issues (Note: Functions properly with  Firefox)

docs/DEV.md
```diff
@@ -0,0 +1,45 @@
...
+
+#### Clone the Repo
+- `git clone https://github.com/scidsg/hushline.git`
+- `cd hushline`
+- `git switch makefile` # Temporary until we merge back into Main
+- `/opt/homebrew/bin/python3 -m venv venv`
+- `source venv/bin/activate`
+- `poetry install`
+- `source env.sh`
+- `poetry run flask db upgrade` _Optional_
```
brassy-endomorph, line 32: This line and the next should be commented with "(Production migration)" and "(Local initialization)" or something similar since they are mutually exclusive and it should be clear why one is picked over the other

hushline/__init__.py
```diff
@@ -35,12 +32,14 @@ def handle_rate_limit_exceeded(e):
     app.config["SESSION_COOKIE_SAMESITE"] = "Lax"
     app.config["PERMANENT_SESSION_LIFETIME"] = timedelta(minutes=30)
 
-    ssl_cert = "/etc/mariadb/ssl/fullchain.pem"
-    ssl_key = "/etc/mariadb/ssl/privkey.pem"
+    # Conditional SSL configuration based on environment
+    if os.getenv("FLASK_ENV") == "production":
```
brassy-endomorph, line 36: Maria DB code sneaking back in

hushline/ext.py
```diff
@@ -1,11 +1,16 @@
+import os
+
 from flask_bcrypt import Bcrypt
 from flask_limiter import Limiter
 from flask_limiter.util import get_remote_address
 
 bcrypt = Bcrypt()
 
+# Check for an environment variable or use in-memory storage as fallback
+REDIS_URI = os.getenv("REDIS_URI")
```
brassy-endomorph, line 10: Global state sneaking back in again. Move it to the `Limiter` initializer

hushline/routes.py
```diff
@@ -232,8 +228,8 @@ def register() -> Response | str:
                 flash("💔 Username already taken.", "error")
                 return redirect(url_for("register"))
 
-            password_hash = bcrypt.generate_password_hash(password).decode("utf-8")
-            new_user = User(primary_username=username, password_hash=password_hash)
+            new_user = User(primary_username=username)
```
brassy-endomorph, line 231: From a previous thread, instead of setting this separately, override the `__init__` method and let `password_hash` be a keyword arg

brassy-endomorph (changes requested):

hushline/__init__.py
glenn-sorrentino: Yeah, I'm going to address that in a separate PR.

hushline/routes.py
glenn-sorrentino: Got it.

docs/DEV.md
glenn-sorrentino: Going to resolve this for now, as I think we should have a sprint that specifically looks at our developer documentation. 

The instructions' current state is what a few of us have gotten to work when setting up the environment. Other deviations from this path have failed. So, given that, they can probably be refined, but they are working and sufficient for the stage we're in. Closing the ticket to continue progress on other efforts.

scidsg-dev (approved):
The requested changes are tested and working.

brassy-endomorph (approved):

Issue #319, opened 2024-04-04
Invalid PGP key causes app to crash. 


glenn-sorrentino:
```
2024-04-04 13:28:32,464:DEBUG:Attempting to validate key: asdf
2024-04-04 13:28:32,469:ERROR:Error validating PGP key: unexpected EOF
2024-04-04 13:28:32,470:ERROR:Error: Could not build url for endpoint 'settings'. Did you mean 'settings.index' instead?
Traceback (most recent call last):
  File "/Users/glennsorrentino/Nextcloud/Git/hushline/venv/lib/python3.12/site-packages/flask/app.py", line 870, in full_dispatch_request
    rv = self.dispatch_request()
         ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/glennsorrentino/Nextcloud/Git/hushline/venv/lib/python3.12/site-packages/flask/app.py", line 855, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)  # type: ignore[no-any-return]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/glennsorrentino/Nextcloud/Git/hushline/venv/lib/python3.12/site-packages/flask_limiter/extension.py", line 1299, in __inner
    return cast(R, flask.current_app.ensure_sync(obj)(*a, **k))
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/glennsorrentino/Nextcloud/Git/hushline/hushline/utils.py", line 20, in decorated_function
    return f(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^
  File "/Users/glennsorrentino/Nextcloud/Git/hushline/hushline/settings.py", line 446, in update_pgp_key
    return redirect(url_for("settings"))
                    ^^^^^^^^^^^^^^^^^^^
  File "/Users/glennsorrentino/Nextcloud/Git/hushline/venv/lib/python3.12/site-packages/flask/helpers.py", line 220, in url_for
    return current_app.url_for(
           ^^^^^^^^^^^^^^^^^^^^
  File "/Users/glennsorrentino/Nextcloud/Git/hushline/venv/lib/python3.12/site-packages/flask/app.py", line 1074, in url_for
    return self.handle_url_build_error(error, endpoint, values)
[...]

Issue #322, opened 2024-04-08
PGP Key Length Bug


glenn-sorrentino:
Fixed in the beta build. Reported on Mastodon that a long (~68k char) PGP key crashed the app. 

Updates:
- https://github.com/scidsg/hushline/blob/main/hushline/settings.py#L54
   - Changed the length limit from 20,000 to 100,000.
- https://github.com/scidsg/hushline/blob/main/hushline/model.py#L20
   - Changed from db.Text to MEDIUMTEXT

PR #323, opened 2024-04-10
Settings Bug Fixes

glenn-sorrentino:
- Increase PGP Key Length to 100,000 characters. 
- Fix fail-state when an invalid key is added.


glenn-sorrentino:
This is tested and working locally.

glenn-sorrentino:
Closes #319 and #322

ritik2311 (approved):
tested it out locally with different keys (valid and invalid)

ritik2311 (approved):

brassy-endomorph (approved):

Issue #276, opened 2024-03-19
Use stronger algorithm for hashing passwords


brassy-endomorph:
We may want to use `scrypt` or `argon2i` so we can control the memory hardness or number of threads to hash a password as a means to increase the difficulty of cracking passwords should the DB ever get leaked.

brassy-endomorph:
Reopening this because `passlib` just wraps and uses `bcrypt`. It gives us a path forward to to updating the password hashes and migrating, but it doesn't actually change the underlying hashing function.

glenn-sorrentino:
Do you have a preference for algorithms? Initial research for Argon2, aside from them winning the 2015 password hashing comp, its newer tech and uses more memory. Scrypt is faster and stronger against brute force.

glenn-sorrentino:
I actually noticed that you recommended Argon2i. Is a middle ground 2id?

glenn-sorrentino:
Working in the `hashing` branch: https://github.com/scidsg/hushline/blob/hashing/hushline/model.py

brassy-endomorph:
After our out-of-band conversation, I think we should pause this so that we can figure out what the appropriate level of compute should, pending threat modeling. If we're not terribly concerned with APT threats, we can use something less complex.

Depends on #265

PR #335, opened 2024-04-25
Add `passlib` and `scrypt`

glenn-sorrentino:
Updated routes to use Passlib and Scrypt:
`/registration`
`/login`
`/change-password`


glenn-sorrentino:
Closes #276

PR #363, opened 2024-06-05
Switch to ruff and mypy for strict linting

micahflee:
In this PR I have deleted black, flake8, and isort, and replaced them all with [Ruff](https://docs.astral.sh/ruff/), which is a single fast linter that does everything those other ones did. I also went through the entire codebase and fixed all the lint issues that ruff found, making the code much nicer.

Previously, mypy had been set up for static type checking, but it hadn't been enforced, and large portions of the code were not actually defining their types. I've started enforcing static types using mypy (which helps find potential bugs before releasing code), and I've also gone through the entire codebase and added type definitions to everything.

I cleaned up the GitHub Actions workflows so that there's just a single workflow that runs the linter and the tests now, instead of two, and I made it so that it strictly enforces ruff and mypy -- basically, GitHub PRs that don't have nicely formatted code and type definitions will be prevented from getting merged. (Honestly, this is something I should do with OnionShare too.)

With all this new stuff, I also updated `DEV.md` to include instructions for the recommended development environment:

> Hush Line is written in Python. To ensure code integrity and consistency, we use [Ruff](https://docs.astral.sh/ruff/) for linting and [mypy](https://www.mypy-lang.org/) for static type checking.
> 
> The recommended development environment is [Visual Studio Code](https://code.visualstudio.com/) with the following extensions:
> 
> - [Python](https://marketplace.visualstudio.com/items?itemName=ms-python.python)
> - [Ruff](https://marketplace.visualstudio.com/items?itemName=charliermarsh.ruff)
> - [Mypy](https://marketplace.visualstudio.com/items?itemName=matangover.mypy)

Using these extensions is cool because it basically shows you all of the ruff and mypy problems directly in VS Code, and helps you fix them, before you even make any git commits.

I also removed a key bit of code that was getting in the way of debugging:

```py
    @app.errorhandler(Exception)
    def handle_exception(e) -> Tuple[str, int]:
        # Consider adjusting error handling as per your logging preferences
        app.logger.error(
            f"Error: {e}", exc_info=True
        )  # Ensure appropriate logging configuration is in place
        return "An internal server error occurred", 500
```

This was capturing exceptions and displaying a generic error, and not printing them to the console. This is why when we loaded `/directory` and it returned error 500, it didn't actually tell you the exception was and why it crashed. Now it does. It's good to hide these details from users in prod, but the correct way of doing that is to run Flask in prod mode instead of in dev mode.

Also, I've switched things around a bit in the Makefile, and the instructions in DEV.md. It basically shows instructions for running Hush Line directly on your computer and not in Docker again. I'll work on finishing up containerizing it in future PRs.


glenn-sorrentino:
This looks great! The only issue I encountered with the updated instructions was running `poetry run migrate`, which resulted in `Command not found: migrate`. However, the app started without issue.

Should the command be `poetry run flask db migrate`?

micahflee:
Oops that should be `poetry run make migrate`.

glenn-sorrentino (approved):
🔥

PR #365, opened 2024-06-10
Database revamp and switch to Postgres in docker

micahflee:
I've deleted the whole migration folder and regenerated it from scratch, with just a single initial migration. I'm not sure if this was strictly necessary or not, but it doesn't hurt.

In #363 I had actually commented out parts of `migrations/env.py` because I couldn't get it working with mypy. So this time instead of trying to fight mypy, I just configured mypy and to ignore the migrations files. I also made ruff ignore one issue in the migrations too, just so they work when generated, without needing to modify them to satisfy the linter each time.

I also made the Docker Compose version of the app use a PostgreSQL database instead of SQLite. You can run this to start the containers:

```sh
docker-compose up --build
```

This now starts 3 containers: `redis`, `postgres`, and the Hush Line `app` itself. `postgres` stores its data in `volumes/postgres`. So now, if we wanted to spin up multiple `app` containers to scale, it should work fine, and they'll all be connecting to the same postgres database.

The postgres database listens on 127.0.01:5432, so you can connect to it there to run SQL queries. If you want to be fancy, you can use a graphical SQL client like this:

Screenshot 2024-06-10 at 3 34 28 PM


jeremywmoore:
@micahflee what impact does the introduction of postgres have to the currently running beta? Presumably it's no longer deployable until we have a production postgres db?

micahflee:
> @micahflee what impact does the introduction of postgres have to the currently running beta? Presumably it's no longer deployable until we have a production postgres db?

It shouldn't have any impact. It just looks at `SQLALCHEMY_DATABASE_URI` , and if it starts with `sqlite://` it uses sqlite, and if it starts with `postgresql+psycopg://` it uses postgres. I didn't actually have to change much to support this except install a new dependency for postgres support, since SQLAlchemy natively supports both.

But that said, this is a good question, and I'm not really sure what changes here or in my last PR might break things in the currently-deployed beta.

PR #366, opened 2024-06-11
Automatically publish container images to GitHub Container Registry

micahflee:
There are various ways to deploy on Digital Ocean App Platform, but the simplest seems to be from a container image hosted on a container registry. This PR adds a GitHub Action workflow to automatically build new container images on every push to the `main` branch, and publish them to the GitHub Container Registry.

This works, but right now the package visibility is private and it can only be accessed with authentication. Making it public would be simpler for deployment (I don't need to store GHCR credentials in Digital Ocean) and it also means others will be able to use the container image that gets built here.

@glenn-sorrentino can you help make it public? Here's the package settings: https://github.com/orgs/scidsg/packages/container/hushline%2Fhushline/settings. Can you scroll down to the Danger Zone at the bottom and click "Change visibility" and then make it public?

image

Once this is public, I can make DO App Platform pull from the container image. Every time any PR is merged into `main`, the dev server will automatically switch to the latest container image.

### DIFF
--- a/.github/workflows/publish_container.yml
+++ b/.github/workflows/publish_container.yml
@@ -1,0 +1,32 @@
+name: Publish container image
+
+on:
+  push:
+    branches:
+      - main
+      - ghcr
+
+jobs:
+  build-and-push:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v4
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Log in to GitHub Container Registry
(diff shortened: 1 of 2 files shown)


jeremywmoore (approved):

glenn-sorrentino (approved):

glenn-sorrentino:
Done!
> @glenn-sorrentino can you help make it public? Here's the package settings: https://github.com/orgs/scidsg/packages/container/hushline%2Fhushline/settings. Can you scroll down to the Danger Zone at the bottom and click "Change visibility" and then make it public?

PR #368, opened 2024-06-13
Doigital Ocean app platform changes

micahflee:
This contains changes required to deploy to DigitalOcean.

This adds a dedicated route called a health check that the hosting platform can load to confirm that the service is up and running. `GET /health.json` responds with `{"status": "ok"}`.

Also making other changes...

### DIFF
--- a/.github/workflows/publish_container.yml
+++ b/.github/workflows/publish_container.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - health-check
 
 jobs:
   build-and-push:
--- a/hushline/__init__.py
+++ b/hushline/__init__.py
@@ -33,15 +33,6 @@ def create_app() -> Flask:
     app.config["SESSION_COOKIE_SAMESITE"] = "Lax"
     app.config["PERMANENT_SESSION_LIFETIME"] = timedelta(minutes=30)
 
-    # Conditional SSL configuration based on environment
-    if os.getenv("FLASK_ENV") == "production":
-        ssl_cert = os.getenv("SSL_CERT_PATH")
-        ssl_key = os.getenv("SSL_KEY_PATH")
-
-        # Ensure SSL files exist
-        if not all(os.path.exists(path) for path in [ssl_cert, ssl_key] if path):
-            raise FileNotFoundError("SSL certificate or key file is missing.")
-
     # Run migrations
     db.init_app(app)
     _ = Migrate(app, db)
--- a/hushline/__init__.py
+++ b/hushline/__init__.py
@@ -3,7 +3,6 @@ import os
 from datetime import timedelta
 from typing import Any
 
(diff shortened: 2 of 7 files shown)


glenn-sorrentino:
I'm okay with removing Flask-Limiter and Redis until we know we need it.

jeremywmoore:
Should we revisit restoring the redis DB and limiter since we no longer think that was the deployment blocker?

glenn-sorrentino:
The goal was app-side DOS mitigation. If there's another more reliable or standard method - Cloudflare or Digital Ocean or something else - I'm open to whatever you think is best. Here's a comment from Brassy regarding Flask-Limiter:

> Yes, for somethings like heavy DB queries or uploads, but the main issue is that someone can just send a bajillion TCP SYNs and then leave the connections open untl the kernel shits the bed. Or other similar things lower in the TCP/IP stack... well before it gets to the application or evern reverse proxy, I mean

start.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+# Run migrations
+poetry run flask db upgrade
+
+# Start the server
+poetry run flask run -p 8080 --host 0.0.0.0
```
jeremywmoore, line 7: this runs the app using flask's development server, we probably want to run via gunicorn?

jeremywmoore (changes requested):
I don't love bundling the removal of the rate limiter with the other changes in this PR but if we think it's not needed we can remove it now. Left some inline comments

.github/workflows/publish_container.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - health-check
```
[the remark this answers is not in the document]
micahflee: Oops. This PR branch was originally called health-check, because I was adding a health-check, and so I wanted to trigger the build in this branch, and I forgot to remove the branch from the GH actions workflow. But also, the DO deploy builds its own container now anyway.

start.sh
micahflee: Good idea

hushline/__init__.py
```diff
@@ -3,28 +3,22 @@
...
-from flask_migrate import Migrate, upgrade
+from flask_migrate import Migrate
 from werkzeug.wrappers.response import Response
 
 from . import admin, routes, settings
 from .db import db
 from .limiter import limiter
 from .model import User
 
-load_dotenv("/etc/hushline/hushline.conf")
```
jeremywmoore, line 17: is this no longer needed because the environment variables are actually provided in the environment?

jeremywmoore (approved):

hushline/__init__.py
micahflee: Yes. Also, it wasn't actually even being used before, as there was no `/etc/hushline/hushline.conf` to load.

PR #369, opened 2024-06-17
Remove flask-limiter and redis

micahflee:
I think it may be a good idea to remove Flask-Limiter and Redis from the app until we're confident it solves a problem we actually have.


pyproject.toml
```diff
@@ -10,19 +10,16 @@ readme = "README.md"
 python = "^3.11"
 cryptography = "^42.0.5"
 flask = "^3.0.2"
-flask-limiter = { extras = ["redis"], version = "^3.7.0" }
 flask-migrate = "^4.0.7"
 flask-sqlalchemy = "^3.1.1"
 flask-wtf = "^1.2.1"
-gunicorn = "^22.0.0"
```
jeremywmoore, line 17: I don't think we want to remove gunicorn

jeremywmoore (commented):
Looks good except for removing the gunicorn dependency, was that intentional?

pyproject.toml
micahflee: yeah, that was not intentional

micahflee:
After merging #368, I'm merged main into this branch, and also added back gunicorn.

pyproject.toml
jeremywmoore: I see the merge commit but diff still shows this as removed

jeremywmoore (approved):

PR #371, opened 2024-06-18
allow specifying the cookie name for localdev

jeremywmoore:
When running locally, Chrome doesn't allow setting cookies that begin with `__Host-` against localhost. This change allows specifying the session cookie name via environment variable for development purposes.  This also reverts to running the debug (vs gunicorn) when running locally.


micahflee:
If the reason local development doesn't work in Chrome is because the cookie is named `__Host-session`, why not just rename the cookie to `session` instead?

jeremywmoore:
@micahflee in this case the `__Host-` prefix acts as a validation on the cookie:
> The cookie must be set with the [Secure attribute](https://owasp.org/www-project-web-security-testing-guide/v41/4-Web_Application_Security_Testing/06-Session_Management_Testing/02-Testing_for_Cookies_Attributes#secure-attribute).
The cookie must be set from a URI considered secure by the user agent.
Sent only to the host who set the cookie and MUST NOT include any [Domain attribute](https://owasp.org/www-project-web-security-testing-guide/v41/4-Web_Application_Security_Testing/06-Session_Management_Testing/02-Testing_for_Cookies_Attributes#domain-attribute).
The cookie must be set with the [Pathattribute](https://owasp.org/www-project-web-security-testing-guide/v41/4-Web_Application_Security_Testing/06-Session_Management_Testing/02-Testing_for_Cookies_Attributes#path-attribute) with a value of / so it would be sent to every request to the host.

Basically it's relevant validation to have on by default for the session cookie, as if any of these requirement aren't met, the browser won't send the cookie.

micahflee:
Ahh okay that makes sense.

micahflee (approved):

PR #376, opened 2024-06-20
Security fixes

micahflee:
A few weeks ago, @lsd-cat reported vulnerabilities in Hush Line, the most serious of which is a persistent stored XSS in the inbox. This PR fixes them all, and also adds tests for 2FA authentication.

Since Hush Line hasn't yet had its first release and it shouldn't be used in production yet, it's fine to make this PR public. Once there's a release cycle, security issues should remain private until fixed.

From the report:

## Stored XSS

> There is a stored XSS in the tip submission form. The input is displayed using the safe Jinja2 attribute, and thus not sanitized upon display. If GPG encryption is enabled, it is trivial to bypass server side encryption by setting `client_side_encrypted=true` in the request and have the plaintext XSS execute in the recipient’s inbox anyway.
> https://github.com/scidsg/hushline/blob/f17285064e51c5f498ddf87f19d7dcf18361f607/hushline/templates/inbox.html#L8
> https://github.com/scidsg/hushline/blob/f17285064e51c5f498ddf87f19d7dcf18361f607/hushline/routes.py#L146

**Solution:**

I fixed it by removing the `| safe` attribute:

https://github.com/scidsg/hushline/blob/1b68f759c8a8c01a368e342da3a1430142d1c085/hushline/templates/inbox.html#L7-L9

## CSP Bypass

> The following CSP directive completely hinders any security property of the policy. Since anybody can upload content to unpkg without any check, than any attacker can too.
> ```
> script-src 'self' https://js.stripe.com https://unpkg.com;
> ```
> There is already a pre-made package for demo purpose. This applies to any similar CDN.
> ```
> <script src="https://unpkg.com/csp-bypass@1.0.2/dist/sval-classic.js"></script>
> <br csp="alert(1)">
> ```
> https://github.com/scidsg/hushline/blob/f17285064e51c5f498ddf87f19d7dcf18361f607/SECURITY.md?plain=1#L49C3-L49C63
> 
> hushline

**Solution:**

I fixed this by vendoring OpenPGP.js and loading it locally instead of from a CDN, and removing `unpkg.com` from the CSP header: https://github.com/scidsg/hushline/blob/2bbeae78a24ca2cd893f32a1812f5f6634cb21b6/hushline/static/vendor/openpgp-5.11.1.min.js

https://github.com/scidsg/hushline/blob/2bbeae78a24ca2cd893f32a1812f5f6634cb21b6/hushline/templates/submit_message.html#L43

## Chain->Plaintext submission leakage/encryption key swap/2FA disable

> Since all tips are displayed on the same page, an attacker can send the following payload to exfiltrate all the tips in the inbox. It can also be exploited to script the change user settings, such as the GPG key in the profile to have any new submission encrypted with an attacker controlled key or to disable 2 factor authentication, since doing so requires neither an OTP nor the user password.

**Solution:**

These are only possible with XSS, and since the XSS is fixed this is not exploitable.

However, requiring a password to disable 2FA is a good idea. Requiring a password to update the PGP key maybe is too, though I think this is more of a usability tradeoff. I haven't implemented these in this PR.

## Unsafe OTP practices

> It seems like there is no protection neither against TOTP reuse nor TOTP bruteforce. A valid TOTP value can be used an unlimited number of times during its time window. Also, after logging in via username and password, there are unlimited attempts (within the generic rate limiting of the application) to guess it. Given there are just ~2^20 possibilites (1 million) and the birthday paradox, it is possible to bypass it with just brute force in a few days (or way less with some luck). The attempts can be made resuing always the same session without the need to re-login every time.
> 
> It is my undertsanding that flask-limiter is ip address based, so in general not a valid countermeasure here. How does it even work when using an Onion address, since every remote address is 127.0.0.1 when dealing with internal Tor connections?

**Solution:**

I have added protection against both TOTP reuse and bruteforce. It works like this. I've added an AuthenticationLog model, to keep track of successful and failed login attempts:

https://github.com/scidsg/hushline/blob/2bbeae78a24ca2cd893f32a1812f5f6634cb21b6/hushline/model.py#L145-L162

When logging in with 2FA, the user is rate-limited if either:

- They have successfully logged in within the last 30 seconds
- They have failed to login 5 times within the last 30 seconds

https://github.com/scidsg/hushline/blob/2bbeae78a24ca2cd893f32a1812f5f6634cb21b6/hushline/routes.py#L317-L341

I've also added a new test file, https://github.com/scidsg/hushline/blob/2bbeae78a24ca2cd893f32a1812f5f6634cb21b6/tests/test_2fa.py, which tests 2FA in general, as well as for these rate limits.

As far as flask-limiter, we recently removed this this from the codebase (#369), so it's no longer relevant.


migrations/versions/e1b5fbe3140a_add_authentication_logs.py
```diff
@@ -0,0 +1,35 @@
...
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('authentication_logs',
+    sa.Column('id', sa.Integer(), nullable=False),
+    sa.Column('user_id', sa.Integer(), nullable=False),
+    sa.Column('successful', sa.Boolean(), nullable=False),
+    sa.Column('timestamp', sa.DateTime(), nullable=False),
+    sa.ForeignKeyConstraint(['user_id'], ['users.id'], ),
+    sa.PrimaryKeyConstraint('id')
```
jeremywmoore, line 27: does this table need an index on user_id + timestamp + successful?

lsd-cat:
The `pyotp` documentation[^1], summarizing from the RFC, suggests:


> - Deny replay attacks by rejecting one-time passwords that have been used by the client (this requires storing the most recently authenticated timestamp, OTP, or hash of the OTP in your database, and rejecting the OTP when a match is seen)
> - Throttle (rate limit) brute-force attacks against your application’s login functionality (see RFC 4226, section 7.3)


And in my experience, this is how most applications do. While I do think the fix proposed here could work security-wise now, I can see two reasons for which the documentation-suggested approach would rather be better:
 - It impairs the user to do any further TOTP action always for more than needed (such as, it locks out the user for 30 seconds even when the TOTP has already rotated).
 - If a `window_size` parameter, supported by the `pyotp.verify()` call, is ever added later in order to improve usability, which is relatively common, then the protection might be ineffective again.

In general, I strongly advise to read the documentation of any security-relevant component used, since most development risks and cases are usually covered.

[^1]: https://pyauth.github.io/pyotp/

migrations/versions/e1b5fbe3140a_add_authentication_logs.py
micahflee: Good idea. In https://github.com/scidsg/hushline/pull/376/commits/2fb60ede18fe68b089ac739dad263a5c0cd70cf0 I recreated the migration, and this time added an index.

micahflee:
@lsd-cat I just pushed a commit that now logs successful login OTP codes, and rejects logins that use the same code twice. This does mean that there is a 1/1000000 chance that it will be rejected as a false positive (if the same OTP is generated twice in a row) ... but I think that's fine, the user can just wait 30 seconds and try again.

> What was the original reason for putting `safe` there? Mailvelope compatibility?

I'm not sure!

glenn-sorrentino (approved):

lsd-cat:
I have not fully audited the code but the explanation and the logic seems ok. Can you please disclose the bugs via the Github feature where they were reported?

glenn-sorrentino:
> I have not fully audited the code but the explanation and the logic seems ok. Can you please disclose the bugs via the Github feature where they were reported?

I requested CVEs. Will publish once they're complete.

PR #378, opened 2024-06-25
Build release container images

micahflee:
This updates the GitHub Actions workflow to trigger on

- any commits to the `main` branch
- or any tags that start with `v`

For commits to the `main` branch, it still builds container images and pushes them to `ghcr.io/scidsg/hushline/hushline:latest`.

When a git tag is created that starts with `v` (for example, `v0.1.0`), it builds a container image and pushes to the following tags:

- `ghcr.io/scidsg/hushline/hushline:v0.1.0` (or whatever your release tag is)
- `ghcr.io/scidsg/hushline/hushline:release`

This way, if you want to run Hush Line from the tip of the `main` branch, you pull the container image with the tag `latest`. If you want to run a specific version, you pull with the tag of the same name as the version (like `v0.1.0`). If you want the latest stable version, you pull with the tag `release`.

The prod version of Hush Line will be running `ghcr.io/scidsg/hushline/hushline:release`. Whenever we make a new release, it will deploy the new version.


.github/workflows/publish_container.yml
```diff
@@ -4,28 +4,52 @@ on:
   push:
     branches:
       - main
+    tags:
+      - 'v*'
 
 jobs:
```
jeremywmoore, line 10: Since the build and push steps are the same except for the tags, we could do something like this to set the desired tags to an environment variable. I'm not sure this is simpler though. If you don't like this, we could also consider just splitting the workflow into two separate workflows.
```suggestion
jobs:
  image-tag-latest:
    if: github.ref == 'refs/heads/main'
    run: |
      echo "IMAGE_TAGS=ghcr.io/${{ github.repository }}/hushline:latest" >> $GITHUB_ENV
      
  image-tag-release:
    if: startsWith(github.ref, 'refs/tags/v')
    run: |
      TAGS="ghcr.io/${{ github.repository }}/hushline:${{ github.ref_name }} \n
                   ghcr.io/${{ github.repository }}/hushline:release"
      echo "IMAGE_TAGS<<EOF" >> $GITHUB_ENV
      echo $TAGS >> $GITHUB_ENV
      echo "EOF" >> $GITHUB_ENV
```

.github/workflows/build_container_latest.yml
```diff
@@ -1,31 +1,31 @@
...
     branches:
       - main
+    tags:
+      - 'v*'
 
 jobs:
-  build-and-push:
+  build-and-push-latest:
     runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/main'
```
jeremywmoore, line 13: can we remove this condition?

.github/workflows/build_container_latest.yml
```diff
@@ -1,31 +1,31 @@
-name: Publish container image
+name: Build latest container image
 
 on:
   push:
     branches:
       - main
+    tags:
```
jeremywmoore, line 7: can we remove the tags from the run trigger?

micahflee: Yup, just fixed

jeremywmoore (approved):

PR #394, opened 2024-07-08
Modifications to flash messages

ispykenny:
1. Removed `z-index` property from styles. Edit: Added `z-index: 10` without altering index value on show/hide
2. Using `transform` instead of `top` for smoother transitions.
3. Updated the hierarchy of the messaging element so it's top level for the screen reader to announce first once it appears on the page.

https://github.com/scidsg/hushline/assets/15894356/09eda57c-f6d8-495f-8507-8cf9f96b1736

**Update screenshots:**

Screenshot 2024-07-08 at 2 13 07 PM
Screenshot 2024-07-08 at 2 12 31 PM

### DIFF
--- a/hushline/static/css/style.css
+++ b/hushline/static/css/style.css
@@ -235,7 +235,7 @@ img.empty {
     max-width: 78vw;
     text-align: center;
     opacity: 0;
-    visibility: hidden;
+    z-index: 10;
     animation: fadeInDown 0.5s ease forwards, fadeOutUp 0.5s ease 5s forwards;
 }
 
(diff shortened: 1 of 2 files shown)


glenn-sorrentino:
> 1. Removed `z-index` property from styles.
> 2. Using `transform` instead of `top` for smoother transitions.
> 3. Updated the hierarchy of the messaging element so it's top level for the screen reader to announce first once it appears on the page.
> 
>  hushline-flashnotification.mov

Looking good @ispykenny! Can we have the flash message lower from the top of the viewport rather than the header so that it appears on top of all other visual elements?

ispykenny:
Updated screenshots 

CC: @glenn-sorrentino 

Screenshot 2024-07-08 at 2 13 07 PM
Screenshot 2024-07-08 at 2 12 31 PM

glenn-sorrentino (approved):

PR #400, opened 2024-07-09
Resolved report link visibility

ispykenny:
1. Fixed JS to render Report account
2. Created endpoint to check if session user to conditionally render link

https://github.com/scidsg/hushline/assets/15894356/b7ee6100-6693-4dbd-b87d-4c31e58c76b3

### DIFF
--- a/hushline/static/js/directory.js
+++ b/hushline/static/js/directory.js
@@ -4,6 +4,7 @@ document.addEventListener('DOMContentLoaded', function () {
     const searchInput = document.getElementById('searchInput');
     const clearIcon = document.getElementById('clearIcon');
     let userData = []; // Will hold the user data loaded from JSON
+    let isSessionUser = false
 
     function updatePlaceholder() {
         const activeTab = document.querySelector('.tab.active').getAttribute('data-tab');
@@ -30,6 +31,12 @@ document.addEventListener('DOMContentLoaded', function () {
             .catch(error => console.error('Failed to load user data:', error));
     }
 
+    async function checkIfSessionUser() {
+        const response = await fetch('/directory/get-session-user.json');
+        const {logged_in} = await response.json();
+        isSessionUser = logged_in;
+    }
+
     function filterUsers(query) {
         const tab = document.querySelector('.tab.active').getAttribute('data-tab');
         return userData.filter(user => {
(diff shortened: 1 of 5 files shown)


glenn-sorrentino:
Part of the feature is for the fields to be pre-filled with the account that's being reported. You can see in the URL: http://127.0.0.1:8080/submit_message/admin?prefill=Reported%20user%3A%20admin%0A%0ABio%3A%20Hush%20Line%20Admin%0A%0AReason%3A

It should display as:

Reported user: [username]
Bio: [bio]
Reason:

glenn-sorrentino (changes requested):
Added a comment about pre-filling the form. Also will request @micahflee to review as well since some Python changes are included.

hushline/static/js/directory.js
```diff
@@ -45,6 +52,20 @@ document.addEventListener('DOMContentLoaded', function () {
         return text.replace(regex, '<span class="search-highlight">$1</span>');
     }
 
+
+    function reportUser(username, bio) {
+
+        // Construct the message content with explicit line breaks
+        const messageContent = `Reported user: ${username}\n\nBio: ${bio}\n\nReason:`;
```
ispykenny, line 59: @glenn-sorrentino isn't this what I have per your comment? Still getting a "User not found error"

I'm looking at our `routes.py` route and don't see anything to support this? Unless I'm missing something? (Sorry for the confusion!)


https://github.com/scidsg/hushline/pull/400#issuecomment-2221300342

ispykenny: Or could it be that I don't have an admin user on my local db?

ispykenny: lol created an`admin` account on my local db and now it works. Sorry 😂!

glenn-sorrentino: What I mean is that when a user clicks "Report Account" it should take then to the admin account's submit message page and have the Message textarea pre-populated with info on the account being reported.

glenn-sorrentino: This is the function in the directory.js file:

```
    function reportUser(username, bio) {

        // Construct the message content with explicit line breaks
        const messageContent = `Reported user: ${username}\n\nBio: ${bio}\n\nReason:`;

        // Encode the message content to ensure line breaks and other special characters are correctly handled
        const encodedMessage = encodeURIComponent(messageContent);
        
        // Redirect to the message submission form for the admin with the pre-filled content
        const submissionUrl = `/submit_message/admin?prefill=${encodedMessage}`;
        window.location.href = submissionUrl;
    }
```

glenn-sorrentino:
Tested and working!

glenn-sorrentino (approved):

Issue #402, opened 2024-07-10
Incorrect elements for tabs (Directory)


ispykenny:
**Describe the bug**
Tab elements are actionable and should be in the tab order. We should make these button elements

**To Reproduce**
1. Go to to the directory page
2. Try tabbing on `Verified` or `All`

**Expected behavior**
Should be able to tab and hit enter to select each tab. 

**Screenshots**
![Screenshot 2024-07-09 at 9 16 01 PM](https://github.com/scidsg/hushline/assets/15894356/8a073900-f943-4118-b6fb-9aecb3578d5b)

**Desktop (please complete the following information):**

PR #404, opened 2024-07-10
Fixed accessibility for directory tabs 

ispykenny:
Allowing tabs to be in the tab order and using `button` elements instead

Example below is shown using keyboard (tabbing) only:

https://github.com/scidsg/hushline/assets/15894356/eefb9bbc-b2f5-4449-a5a4-8aae10b801c3


ispykenny:
Issue: https://github.com/scidsg/hushline/issues/402

glenn-sorrentino (approved):

Issue #403, opened 2024-07-10
Don't allow emoji or alternate fonts for username field


glenn-sorrentino:
**Describe the bug**
A user created an account using an emoji with a differently rendered font.

**Expected behavior**
- More rules around usernames
- No emoji
- No alternate fonts
- Character limit

**Screenshots**
![Screenshot 2024-07-09 at 10 28 27 PM](https://github.com/scidsg/hushline/assets/28545431/53df31bf-5477-46fb-9d60-8bbf5edb7c82)

PR #406, opened 2024-07-10
Fix username validation

glenn-sorrentino:
Updated the Regex for username validation to avoid emoji and other non-standard patterns.

**New Rules:**
- Usernames must begin and end with an alphanumeric character.
- Usernames can contain alphanumeric characters, hyphens, and underscores.
- Usernames cannot contain emojis or other non-standard characters.
- Length of the username must be between 4 and 20 characters.

### Regex Pattern:
```regex
^[a-zA-Z0-9]([a-zA-Z0-9_-]{2,18}[a-zA-Z0-9])?$
```

Screenshot 2024-07-10 at 1 29 44 PM
Screenshot 2024-07-10 at 1 30 17 PM
Screenshot 2024-07-10 at 1 34 18 PM

### DIFF
--- a/hushline/routes.py
+++ b/hushline/routes.py
@@ -29,8 +29,8 @@ logging.basicConfig(level=logging.INFO, format="%(asctime)s:%(levelname)s:%(mess
 
 
 def valid_username(form: Form, field: Field) -> None:
-    # Regex pattern with length constraint (3 to 20 characters) and no special characters at the start or end
-    if not re.match(r"^[a-zA-Z0-9][a-zA-Z0-9_-]{4,20}[a-zA-Z0-9]$", field.data):
+    # Corrected regex pattern with length constraint (4 to 20 characters)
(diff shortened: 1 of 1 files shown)


glenn-sorrentino:
Closes #403

hushline/routes.py
```diff
@@ -29,9 +29,12 @@
 
 
 def valid_username(form: Form, field: Field) -> None:
-    if not re.match(r"^[a-zA-Z0-9_-]+$", field.data):
+    # Corrected regex pattern with length constraint (4 to 20 characters)
+    if not re.match(r"^[a-zA-Z0-9]([a-zA-Z0-9_-]{2,18}[a-zA-Z0-9])?$", field.data):
```
jeremywmoore, line 33: I think wrapping the new part of the regex in `(...)?` means that it'll match if there are 0 or 1 instances, meaning that this will also match a single character username. This might be more like what you are looking for:

```suggestion
    if not re.match(r"^[a-zA-Z0-9][a-zA-Z0-9_-]{2,18}[a-zA-Z0-9]$", field.data):
```

I'd definitely recommend writing a test for this function.

PR #410, opened 2024-07-15
Fix Message Debouncing

glenn-sorrentino:
Right now in Hush Line when submitting a message it's possible to click "Send Message" and then a second later click it again, which ends up sending two messages. This fix makes it so that when the user clicks submit and the form starts processing the button becomes disabled.

Added js:
```js
document.addEventListener("DOMContentLoaded", function() {
...
    document.getElementById('messageForm').addEventListener('submit', function() {
        document.getElementById('submitBtn').disabled = true;
    });
});
```


ispykenny:
Tested and verified

ispykenny (approved):

Issue #357, opened 2024-05-28
Epic: Future-Proofing - Sprint 1


glenn-sorrentino:
Source of Truth: [Planning Doc](https://docs.google.com/document/d/1aghE3nkuWr_rCr5HfCKc3E0s7UbDMpcdrlnu8Bwaq68/edit?usp=sharing)

## Time Estimate
X-Large: 45+ Hours

## Overview
As tech advances and threats evolve, Hush Line must adapt to stay up-to-date. Inherent to PGP are longstanding issues—asking users to create and manage public and private keys introduces significant risks to usability and security and can lead to poor adoption. 

We’re excited to investigate ways to future-proof Hush Line, including, but not limited to, rewriting the application in a memory-safe language like RUST and updating our encryption scheme to the [Signal Double-Ratchet Protocol](https://signal.org/docs/specifications/doubleratchet/). This work may span to H2.

Issue #289, opened 2024-03-20
Determine utility of encrypted DB fields


brassy-endomorph:
We need a threat model to determine whether row-level security like following makes sense

https://github.com/scidsg/hushline/blob/2c89aa633d6ee2eaeb4eeb1e76e349e65a648e2d/app.py#L223-L229

My general suspicion is that the model we come up with will not have a meaningful increase in security by this feature and it only adds complexity. We likely want to significantly harden the database and server against exfiltration before we add this in. The app is the only thing interacting with the DB, and RCE is most likely to happen in the app itself (which has access to the decryption keys).

Blocked by #265

glenn-sorrentino:
Closing, as this encryption hasn't caused any issues, and we want to encrypt everything. A ticket for hardening the server against exfil should be created if this is a priority, but at-rest encryption is something I want to keep and use as broadly as possible.

brassy-endomorph:
If you want DB encryption (beyond the drive's FDE), then we likely want to [encrypt the whole database](https://sqlite.org/com/see.html). I think this needs to remain open. "Hasn't caused problems yet" isn't sufficient reason to keep this, and the main threat of exfil will be the application process / user itself being compromised in which case it can just read the entire database, env vars, and config itself.

rmlibre:
Encrypting the whole database is a good idea. It's also a good idea to encrypt fields that hold secrets. That way they don't linger around in RAM for longer than needed. Regardless of the practicality of attacks, it's best practice to wipe secrets & key material from memory ASAP.

Issue #268, opened 2024-03-16
Better encryption key management


glenn-sorrentino:
We're storing the encryption key in environment variables, but a more secure method might be beneficial in case of server compromise. 

Some options:

- AWS Secrets Manager
- HashiCorp Vault 
- Azure Key Vault
- Google Cloud Secret Manager
- CyberArk Conjur
- Docker Secrets
- Bitwarden

brassy-endomorph:
Suggestion to not have this in the milesone [1. Best Practices and Tidying Up](https://github.com/scidsg/hushline/milestone/1) as this is non-trivial and is fairly advanced. It would require us to run another service, and we'd need that service to do signing and decrypting, not just hand the key over to us on request.

rmlibre:
What's the desired administrator experience for authenticating when starting up the server? Is it reasonable to assume the administrator can use a local password manager or a hardware security key? Those would be secure options to access a [key encryption key](https://csrc.nist.gov/glossary/term/Key_Encryption_Key) (KEK) which then unlocks other keys & secrets in the database.




> Some options:
>
> * AWS Secrets Manager
> * HashiCorp Vault
> * Azure Key Vault
> * Google Cloud Secret Manager
> * CyberArk Conjur
> * Docker Secrets
> * Bitwarden

I don't think the security model covers trusting remote third-parties with administrator secrets?

PR #411, opened 2024-07-16
[DRAFT] feat: begin modernization of symmetric crypto [#268][#289][#357]

rmlibre:
# Description

Generally, this PR is aimed at bringing best practices to the handling of secrets. That is done through making them more inaccessible, or by clearing them from memory as soon as possible. This PR also brings domain separation, input canonicalization, context commitment, & authenticated associated data to distinct encryption / decryption contexts.

More details & updates to come. Comments & reviews are open.

## Domain Separation:

When using cryptography to authenticate data, or derive secrets, etc, it's important to tie those processes to their intended usage contexts by declaring the context with an unambiguous _label_. This allows both the production & consumption of cryptographic artifacts to ensure that they're in agreement over the label (domain) the artifact is supposed to be used within.

#### References:
[What is meant by domain separation ...](https://crypto.stackexchange.com/a/66970)
[What non-trivial benefit does including ...](https://crypto.stackexchange.com/a/106180)
[Separate Your Domains: NIST PQC ...](https://eprint.iacr.org/2020/241.pdf)

## Input Canonicalization:

Cryptography works best when its users are 100% sure & clear about what information they're processing. That can be subtlety, and surprisingly, challenging to do.

#### Example:
Say we're not using canonical input encoding, and we're hashing (H) a username: "green" & an email address: "housekeeper[at]email[dot]com". Here, if the delineation is unclear, then the information is not clear. Another user with username: "greenhouse" and email address: "keeper[at]email[dot]com" will produce the same hash.

```python
H("green" + "housekeeper@email.com") == H("greenhouse" + "keeper@email.com")
```

#### References:
[What is Canonicalization Attack?](https://www.geeksforgeeks.org/what-is-canonicalization-attack/)
[Canonicalization Attacks Against ...](https://soatok.blog/2021/07/30/canonicalization-attacks-against-macs-and-signatures/)

## AEAD & Context Commitment:

Confusingly, the two most widely deployed authenticated ciphers (AES-GCM & ChaCha20-Poly1305) do not create ciphertexts which commit to a key. Which means that knowing the key allows one to create a single ciphertext which successfully decrypts to various different plaintext messages. In most cases this isn't considered an issue because it requires knowledge of the key. But, it does violate intuitive notions of what it means for something to be authentic when forgeries are easy to produce. It has also been shown to lead to practical problems:

> Partitioning oracles can arise when encryption schemes are not committing with respect to their keys. We detail adaptive chosen ciphertext attacks that exploit partitioning oracles to efficiently recover passwords and de-anonymize anonymous communications.
>
> ~ Source: [USENIX Security 2021](https://www.usenix.org/conference/usenixsecurity21/presentation/len)

Because Fernet uses [HMAC-SHA-256], instead of the universal hashes of the previously discussed ciphers, it doesn't suffer from this problem. But, it also doesn't have a direct interface for authenticating domain labels or other associated data. Borrowing partially from [Efficient Schemes for Committing ...], which proposes deriving a fresh key by running a keyed PRF over all of the inputs, it can elegantly support the authentication of additional values.

#### References:
[Key Committing AEADs](https://eprint.iacr.org/2020/1153.pdf)
[How to Abuse and Fix Authenticated Encryption ...](https://eprint.iacr.org/2020/1456.pdf)
[Succinctly-Committing Authenticated Encryption](https://eprint.iacr.org/2024/875.pdf)
[Key commitment in GCM (or AEAD in general)](https://crypto.stackexchange.com/a/108203)

# Remediations

... _detailed descriptions coming soon_

---

### Passing Workflows:

- [x] [Run Linters & Tests](https://github.com/rmlibre/hushline/actions/runs/9954290392/job/27499596344) on (https://github.com/scidsg/hushline/pull/411/commits/920311bc263478159c364fa661936d34eb4272d1)
- [x] [Run Linters & Tests](https://github.com/rmlibre/hushline/actions/runs/9955017167/job/27501888069) on (https://github.com/scidsg/hushline/pull/411/commits/160fa0128723c6296148d41c363d56bb66bd5aec)
- [x] [Run Linters & Tests](https://github.com/rmlibre/hushline/actions/runs/9985937780/job/27597624997) on (https://github.com/scidsg/hushline/pull/411/commits/a3b3f25222b47e493f207a13ffe9392823ce300a) got ⚠️ **Secrets Uncovered [Warning](https://github.com/scidsg/hushline/pull/411/checks?check_run_id=27597624750)**
- [x] [Run Linters & Tests](https://github.com/rmlibre/hushline/actions/runs/10001105408/job/27647994003) on (https://github.com/scidsg/hushline/pull/411/commits/f9e2d79359456962fc1733d50c0d7a5e9702a16d)
- [x] [Run Linters & Tests](https://github.com/rmlibre/hushline/actions/runs/10021948873/job/27701172042) on (https://github.com/scidsg/hushline/pull/411/commits/4c71366e1f2d12b9f3d7de389a5b1e9d33880a11)
- [x] [Run Linters & Tests](https://github.com/rmlibre/hushline/actions/runs/10024476009/job/27706597602) on (https://github.com/scidsg/hushline/pull/411/commits/294338bb20d60ed5a2900f842e1fb432460d5d3a)
- [x] [Run Linters & Tests](https://github.com/rmlibre/hushline/actions/runs/10044653045/job/27760176817) on (https://github.com/scidsg/hushline/pull/411/commits/8ca05bac11d28b417a275761e93b648130f135ea)
- [x] [Run Linters & Tests](https://github.com/rmlibre/hushline/actions/runs/10048132721/job/27771421669) on (https://github.com/scidsg/hushline/pull/411/commits/6f53d35d4a1c4d29292f3bb7c093187192a02ea9)
- [x] [Run Linters & Tests](https://github.com/rmlibre/hushline/actions/runs/10105117734/job/27945136097) on (https://github.com/scidsg/hushline/pull/411/commits/0924906c437c77a2c3f89a401cef0c2b2d4d5016)
- [x] [Run Linters & Tests](https://github.com/rmlibre/hushline/actions/runs/10105258529/job/27945457099) on (https://github.com/scidsg/hushline/pull/411/commits/d6af3e3d2b457c5b8023f99b4d0550164c4603dc)
- [x] [Run Linters & Tests](https://github.com/rmlibre/hushline/actions/runs/10116454546/job/27979337524) on (https://github.com/scidsg/hushline/pull/411/commits/c51e6e8f2590e5a9c5e865d0730b42ba6a28159d)
- [x] [Run Linters & Tests](https://github.com/rmlibre/hushline/actions/runs/10119818937/job/27988783626) on (https://github.com/scidsg/hushline/pull/411/commits/7799379088e833282a0a8b9e8f9fd21756321522)
- [x] [Run Linters & Tests](https://github.com/rmlibre/hushline/actions/runs/10164767244/job/28110969877) on (https://github.com/scidsg/hushline/pull/411/commits/ee93c6a77070f153b484a25fcecd9123d631f0a4)

[//]: # (in-lined links)

[HMAC-SHA-256]: https://github.com/fernet/spec/blob/master/Spec.md#hmac
[Efficient Schemes for Committing ...]: https://eprint.iacr.org/2022/268.pdf


hushline/crypto/secrets_manager.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def _prepare_key_derivation_object(self, admin_secret: bytearray) -> None:
+        salt = self._summon_device_salt()
+        hashed_secret_with_metadata = bytearray(
+            argon2
+            .using(salt=bytes(salt), digest_size=self._KDF_BLOCKSIZE, memory_cost=self._memory_cost)
+            .hash(bytes(admin_secret)),
+            encoding="utf-8",
+        )
+        hashed_secret = truncated_b64decode(hashed_secret_with_metadata.split(b"$")[-1])
```
jeremywmoore, line 72: what's the purpose of this truncated `hashed_secret` value vs initializing the hash with the hashed secret+metadata?

hushline/crypto/secrets_manager.py
```diff
@@ -0,0 +1,113 @@
...
+        self._kdf = shake_256(hashed_secret)
+        self._kdf.update(fullblock_ljust(hashed_secret_with_metadata, self._KDF_BLOCKSIZE))
+        salt.clear()
+        admin_secret.clear()
+        hashed_secret.clear()
+        hashed_secret_with_metadata.clear()
+
+    def derive_key(
```
jeremywmoore, line 80:
```suggestion
    def _derive_key(
```
is this function intended to be exported from the module?

jeremywmoore (commented):
Here's a first pass review. Another pass may yield some more suggestions but I think there are some things to work out around managing the salt.

hushline/crypto/secrets_manager.py
rmlibre: This is a fine detail regarding how to turn a shake_256 object into a keyed permutation. About a year ago I wrote about a more in-depth security [analysis](https://crypto.stackexchange.com/a/107513) of keyed SHA-3 objects. But essentially, the reason is because it's easier to make security arguments when using [_uniformly random_](https://en.wikipedia.org/wiki/Discrete_uniform_distribution) keys. But argon2 output looks like this:

```bash
'$argon2id$v=19$m=65536,t=3,p=4$ACAkhBCCkLL2XotxrnXOuQ$AW3E2OwuNEId...'
```

This metadata, and the encoding, are not uniformly random. So, only the raw, random hash part of the output is used as the key to initialize the keyed internal state. After it's initialized, the whole argon2 output is ingested as a commitment to the encoded original, and then padded to a multiple of the internal block-size.

https://github.com/scidsg/hushline/blob/160fa0128723c6296148d41c363d56bb66bd5aec/hushline/crypto/secrets_manager.py#L74


Here's a bit of a visual to see what's going on:


```bash
'''
# the initial shake_256 state is 200 bytes of zeros
"00000000000000000000000000000000000000000000000000"
                 ⊕
# the raw hashed secret is exactly the size of the internal block-size
"hashedsecrethashedsecrethashedsecr"
                 =
# the raw hashed secret gets XORd into the state
"hashedsecrethashedsecrethashedsecr0000000000000000"

# then, scrambling occurs after a block-size number of bytes are ingested
π("hashedsecrethashedsecrethashedsecr0000000000000000")
                 =
# after the scrambling, the whole internal state is randomized
"0ih7zekm1l5cea6i6yy5bx54bk9pd4e9k3u4zpgbpah675dd5m"
'''
```




---






Another visual of the internal [SHA-3](https://en.wikipedia.org/wiki/SHA-3) "sponge":

[Sponge-Construction-Keccak-000](https://keccak.team/sponge_duplex.html)

hushline/crypto/secrets_manager.py
```diff
@@ -0,0 +1,113 @@
...
+def truncated_b64decode(value: bytearray) -> bytearray:
+    if bool(remainder := len(value) % 4):
+        value.extend((4 - remainder) * b"=")
+    return bytearray(b64decode(value))
+
+
+class SecretsManager:
+    __slots__ = ("_secret_salt_filename", "_secret_salt_length", "_memory_cost", "_kdf")
+
+    _APP_SECRETS_DIRECTORY: Path = Path("hushline/crypto/_app_secrets")
```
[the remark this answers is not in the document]
rmlibre: That's a good idea. Is ``/var/run/...`` available on all platforms? I'm not certain, but perhaps a private hushline home folder directory is more cross-platform? 

i.e something like:
```python
Path("~/.hushline/_app_secrets").expanduser()
```

hushline/crypto/secrets_manager.py
```diff
@@ -0,0 +1,113 @@
...
+        self._secret_salt_length = (
+            self._SECRET_SALT_LENGTH if secret_salt_length is None else secret_salt_length
+        )
+        self._prepare_key_derivation_object(admin_secret)
+
+    def _summon_device_salt(self) -> bytearray:
+        salt_path = self._APP_SECRETS_DIRECTORY / self._secret_salt_filename
+
+        if not salt_path.is_file():
+            salt_path.write_bytes(salt := token_bytes(self._secret_salt_length))
```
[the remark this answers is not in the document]
rmlibre: Hm, the local filesystem isn't available in some of the intended environments... It would be preferred to keep it out of the database... But, if it's the only commonly available location, it may have to reside there.

@jeremywmoore @micahflee Would there be any technical issues preventing the salt from being stored in the database?

hushline/__init__.py
```diff
@@ -8,18 +10,37 @@
...
+from .crypto import SecretsManager
 from .db import db
 from .model import User
 
 
+def _production_app_secrets_insertion(app: Flask) -> None:
+    app.config["VAULT"] = SecretsManager(
+        bytearray(getpass.getpass("admin secret: "), encoding="utf-8")
+    )
+    app.config["SECRET_KEY"] = os.environ.get("SECRET_KEY")
```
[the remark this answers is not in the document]
rmlibre: Would you mind clarifying what you mean by this point? Do you mean line 22?

hushline/crypto/secrets_manager.py
rmlibre: That's a good point. It's best to keep it internal.

Addressed by (https://github.com/scidsg/hushline/pull/411/commits/5407394eb77d5baa6497fa3ef9d83aaa8e70034f)

hushline/__init__.py
```diff
@@ -8,18 +10,37 @@
...
+
+
+def _development_app_secrets_insertion(app: Flask) -> None:
```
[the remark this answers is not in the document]
rmlibre: Adopted by (https://github.com/scidsg/hushline/pull/411/commits/5407394eb77d5baa6497fa3ef9d83aaa8e70034f)

hushline/__init__.py
```diff
@@ -8,18 +10,37 @@
 from werkzeug.wrappers.response import Response
 
 from . import admin, routes, settings
+from .crypto import SecretsManager
 from .db import db
 from .model import User
 
 
+def _production_app_secrets_insertion(app: Flask) -> None:
```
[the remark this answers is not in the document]
rmlibre: Adopted by (https://github.com/scidsg/hushline/pull/411/commits/5407394eb77d5baa6497fa3ef9d83aaa8e70034f)

hushline/model.py
```diff
@@ -54,57 +54,57 @@ def check_password(self, plaintext_password: str) -> bool:
 
     @property
     def totp_secret(self) -> str | None:
-        return decrypt_field(self._totp_secret)
+        return decrypt_field(self._totp_secret, domain=b"totp_secret")
```
rmlibre, line 57: @micahflee @jeremywmoore These row encryptions aren't making use of the ``aad`` kwarg yet. I need more insight from the team to around which value(s) could be used as a unique / canonical representation of a user. Best if it doesn't ever change for the duration of the user's admittance to the database. If it does change, then all the encrypted values will have to be re-encrypted under the new context. Possible ideas: email address, user-id, public pgp key...

hushline/__init__.py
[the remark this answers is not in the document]
jeremywmoore: exactly, since we narrowed the scope of the function name and `SECRET_KEY` is always derived from the environment, let's extract this line into the caller.

hushline/crypto/secrets_manager.py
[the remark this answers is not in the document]
jeremywmoore: `/var/run/` should be available on linux/unix systems (including mac os), https://www.reddit.com/r/linux/comments/cpisy/comment/c0ua3mo/

hushline/crypto/secrets_manager.py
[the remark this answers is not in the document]
jeremywmoore: I don't see a technical reason why we couldn't.

hushline/crypto/secrets_manager.py
jeremywmoore: This seems like a great reason, let's just add a comment to that effect, something like "The initial key of the shake_256() should be uniformly random, so strip the metadata and encoding from the hashed secret to use as our initial key"

hushline/model.py
jeremywmoore: I'd love for @glenn-sorrentino to weigh in on this one as it feels like a product question :)

jeremywmoore (commented):
How does having a single shared salt compare to generating a per-user salt? Does using a kdf which incorporates the domain (and possible aad) obviate this need? 

Although since we may still want to generate a per-user salt for passwords to combat precompute attacks.

hushline/crypto/secrets_manager.py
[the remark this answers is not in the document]
rmlibre: Since saving to the local filesystem & sharing that data between instances isn't universally available, I'll be ripping out the filesystem logic. I'll pivot towards a database approach. Perhaps a new model like ``InfrastructureAdmin``?

hushline/crypto/secrets_manager.py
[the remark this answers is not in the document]
rmlibre: Okay, I'll pivot towards using the database for initialization values.

hushline/__init__.py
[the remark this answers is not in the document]
rmlibre: Ok. I'm moving towards storing this encrypted in the database, to be retrieved during initialization. But, I think the logic will be shared for all flows, so I agree that it should be pulled out from these functions.

rmlibre:
@jeremywmoore:
> How does having a single shared salt compare to generating a per-user salt? Does using a kdf which incorporates the domain (and possible aad) obviate this need?
> 
> Although since we may still want to generate a per-user salt for passwords to combat precompute attacks.


A per-user unique secret is a great idea ✅.

They're nice with regard to the ``aad`` kwarg in field encryptions, because they can be dedicated for that purpose, relieving pressure from other user values. And, they do so while being ideal high-entropy values.

In the same vein, it'd be best to avoid overloading their purpose in ways which replace other dedicated values, like the admin secrets, or the salts automatically generated by a password hashing algorithm. Somehow, where applicable, utilizing them as additional supporting values may be okay.

hushline/crypto/secrets_manager.py
rmlibre: Addressed by (https://github.com/scidsg/hushline/pull/411/commits/4dd848812843242de2959eed0afdec89b6cc16a5)

hushline/crypto/secrets_manager.py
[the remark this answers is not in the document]
rmlibre: Addressed by (https://github.com/scidsg/hushline/pull/411/commits/294338bb20d60ed5a2900f842e1fb432460d5d3a)

hushline/crypto/secrets_manager.py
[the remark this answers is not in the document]
rmlibre: Addressed by (https://github.com/scidsg/hushline/pull/411/commits/294338bb20d60ed5a2900f842e1fb432460d5d3a)

hushline/__init__.py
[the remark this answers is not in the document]
rmlibre: Addressed by (https://github.com/scidsg/hushline/pull/411/commits/294338bb20d60ed5a2900f842e1fb432460d5d3a)

hushline/model.py
rmlibre: Addressed by (https://github.com/scidsg/hushline/pull/411/commits/8ca05bac11d28b417a275761e93b648130f135ea)

hushline/model.py
```diff
@@ -39,72 +75,129 @@ class User(Model):
...
+            # user-specific `aad` components, the admin secret, & the other domain separators.
+            # Reference:
+            # https://www.youtube.com/watch?v=OQD3qDYMyYQ
+            #
+            # This 32-byte pre-hash is also hex-normalized, which avoids implementation-specific
+            # hazards in some password hashing algorithms over raw bytes, & too long passwords.
+            # Reference:
+            # https://blog.ircmaxell.com/2015/03/security-issue-combining-bcrypt-with.html
+            pre_hashed_password = vault._derive_key(domain=domain, aad=aad)
+            self._password_hash = scrypt.hash(pre_hashed_password.hex())
```
rmlibre, line 118: Switch from scrypt to the preferred / recommended argon2.

hushline/model.py
```diff
@@ -39,72 +75,129 @@ class User(Model):
...
     def check_password(self, plaintext_password: str) -> bool:
         """Check the plaintext password against the stored hash."""
-        return scrypt.verify(plaintext_password, self._password_hash)
+        domain = b"user_password_hash"
+        vault = current_app.config["VAULT"]
+        with temp_user_aad(self) as aad:
+            aad.append(self._password_revision_number.to_bytes(8, "big"))
+            aad.append(bytearray(plaintext_password, encoding="utf-8"))
+            pre_hashed_password = vault._derive_key(domain=domain, aad=aad)
+        return scrypt.verify(pre_hashed_password.hex(), self.password_hash)
```
rmlibre, line 128: Switch from scrypt to the preferred / recommended argon2.

hushline/settings.py
```diff
@@ -177,9 +176,9 @@ def index() -> str | Response:  # noqa: PLR0911, PLR0912
 
             # Handle Change Password Form Submission
             if change_password_form.validate_on_submit():
-                if scrypt.verify(change_password_form.old_password.data, user.password_hash):
+                if user.check_password(change_password_form.old_password.data):
                     # Hash the new password using scrypt and update the user object
```
rmlibre, line 180: Switch from scrypt to the preferred / recommended argon2.

hushline/model.py
```diff
@@ -1,10 +1,13 @@
 import secrets
+from collections import deque
+from contextlib import contextmanager
 from datetime import datetime, timedelta, timezone
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Generator
 
 from flask import current_app
 from flask_sqlalchemy.model import Model
 from passlib.hash import scrypt
```
rmlibre, line 9: Switch from scrypt to the preferred / recommended argon2.

rmlibre (commented):
Find uses of scrypt & swtich to the preferred / recommended argon2.

hushline/model.py
rmlibre: Addressed by (https://github.com/scidsg/hushline/pull/411/commits/6f53d35d4a1c4d29292f3bb7c093187192a02ea9)

hushline/settings.py
rmlibre: Addressed by (https://github.com/scidsg/hushline/pull/411/commits/6f53d35d4a1c4d29292f3bb7c093187192a02ea9)

hushline/model.py
rmlibre: Addressed by (https://github.com/scidsg/hushline/pull/411/commits/6f53d35d4a1c4d29292f3bb7c093187192a02ea9)

hushline/model.py
rmlibre: Addressed by (https://github.com/scidsg/hushline/pull/411/commits/6f53d35d4a1c4d29292f3bb7c093187192a02ea9)

hushline/model.py
```diff
@@ -39,72 +75,129 @@ class User(Model):
...
     def password_hash(self, plaintext_password: str) -> None:
-        """Hash plaintext password using scrypt and store it."""
-        self._password_hash = scrypt.hash(plaintext_password)
+        """Hash plaintext password using argon2 and store it."""
+        domain = b"user_password_hash"
+        vault = current_app.config["VAULT"]
+        self._password_revision_number += 1
+        with temp_user_aad(self) as aad:
+            aad.append(self._password_revision_number.to_bytes(8, "big"))
+            aad.append(bytearray(plaintext_password, encoding="utf-8"))
```
rmlibre, line 104: The encodings here aren't committed to. Also, the revision number could be bigger to allow for algorithm version information to be encoded into a prefix.

hushline/model.py
```diff
@@ -39,72 +75,129 @@ class User(Model):
...
+            self._password_hash = argon2.hash(pre_hashed_password.hex())
 
     def check_password(self, plaintext_password: str) -> bool:
         """Check the plaintext password against the stored hash."""
-        return scrypt.verify(plaintext_password, self._password_hash)
+        domain = b"user_password_hash"
+        vault = current_app.config["VAULT"]
+        with temp_user_aad(self) as aad:
+            aad.append(self._password_revision_number.to_bytes(8, "big"))
+            aad.append(bytearray(plaintext_password, encoding="utf-8"))
```
rmlibre, line 126: The encodings here aren't committed to. Also, the revision number could be bigger to allow for algorithm version information to be encoded into a prefix.

rmlibre (commented):
Commit to the password hash algorithm's associated encodings

hushline/model.py
rmlibre: Addressed by (https://github.com/scidsg/hushline/pull/411/commits/c23c9beb78af1265ec52f56efa9c0740759c7724)

hushline/model.py
rmlibre: Addressed by (https://github.com/scidsg/hushline/pull/411/commits/c23c9beb78af1265ec52f56efa9c0740759c7724)

brassy-endomorph:
Since this is a stale branch, can we close the PR and eventually split this out into separate tasks?

PR #413, opened 2024-07-16
Accessibility feedback improvements 

ispykenny:
Hey all,

Here's a walk-through of the changes included in this PR

https://www.loom.com/share/72e0d673639f4b11bea9111d9ef6c325?sid=386bfe34-29d2-460c-a706-84c369c31cb9

This PR resolves the following:

- [x] - the tabsets aren’t keyboard accessible, just mouse - check out the W3C’s APG for a guide.
- [x] - search input needs a visual label (icon of magnifying glass is plenty)
- [x] - your username dropdown should decide what it wants to be. Either a “disclosure widget” or a “menu button”.  Either way you can get the markup/keyboard from the W3C’s APG.
- [x] - can’t tab to the X in your search b/c it’s just a clickable span. Should be a button
- [x] - use the <mark> tag for your search highlight instead of <span>
- [x] - same idea with directory cards, put them in an article or even a role=“group” and name them (aria-label) with the name of the user. 
- [x] - your cards are just divs. I’d give them <article> tags and give them an aria-label to name/identify them. This way the Delete buttons are programatically associated to something. Otherwise it’s just a bunch of orphaned delete buttons not the page.
- [x] - the toast alerts need role=“alert” added to them for screen readers


glenn-sorrentino (changes requested):
This is looking great! Only nit is the search icon. Can we hide it and still make it work? If we absolutely need it, can we style like this? 
Directory

glenn-sorrentino:
> This is looking great! Only nit is the search icon. Can we hide it and still make it work? If we absolutely need it, can we style like this? Directory

From Jesse:
Screenshot 2024-07-17 at 10 08 08 AM

ispykenny:
Screenshot 2024-07-17 at 12 07 20 PM


Implemented new image icon

glenn-sorrentino (dismissed):

ispykenny:
@glenn-sorrentino if I could get another look through on this. TIA.

glenn-sorrentino:
did you intend to have the directory voive over update included?

PR #432, opened 2024-07-20
use the gha docker build cache

jeremywmoore:
Docker builds currently take ~30 minutes across both architectures. Container builds benefit from caching as the lower layers rarely change.  This PR enables reading from and writing to the github actions.


jeremywmoore:
Here's a concise write-up on different caching options, https://engineering.deptagency.com/how-to-speed-up-docker-builds-in-github-actions.

glenn-sorrentino (approved):

PR #377, opened 2024-06-25
perf(auth): avoid unnecessary 30sec TOTP lockout [#376]

rmlibre:
## Description

A rate limit was imposed if a user had previously authenticated with the same TOTP code that they're currently authenticating with. The [TOTP](https://en.wikipedia.org/wiki/Time-based_one-time_password) [spec](https://www.rfc-editor.org/rfc/rfc6238) doesn't disallow repeat codes across time intervals. Such collisions are to be expected when using a pseudo-random function ([HMAC](https://www.rfc-editor.org/rfc/rfc4226)), particularly when using short outputs. The rate limit imposition can harm UX with high authentication latencies without reasonable justification.

## Expected Behavior

A user can authenticate with the same TOTP code if it's a valid code generated at a distinct time interval from the committed TOTP code stored in the database.

## Remediations

1. ``pyotp``'s [timecode](https://github.com/pyauth/pyotp/blob/249674bf2814e1fcc3152c618514159cd44b4610/src/pyotp/totp.py#L106) can be utilized to assess the current time interval.

2. A timecode can be generated at auth initialization, & committed to the database along with the TOTP code upon successful auth.

https://github.com/scidsg/hushline/blob/4eba1e49120871a13f8a36dea1b58c9c5d45844f/hushline/routes.py#L317-L320

https://github.com/scidsg/hushline/blob/4eba1e49120871a13f8a36dea1b58c9c5d45844f/hushline/routes.py#L355-L358

3. Subsequent auth attempts distinguish between time intervals when assessing problematic TOTP code reuse.

https://github.com/scidsg/hushline/blob/4eba1e49120871a13f8a36dea1b58c9c5d45844f/hushline/routes.py#L331-L339

### Passing Workflows:

- [x] [Run Tests](https://github.com/rmlibre/hushline/actions/runs/9654647542)
- [x] [Publish container image](https://github.com/rmlibre/hushline/actions/runs/9654647560)

PR #434, opened 2024-07-20
'Revision' make target

jeremywmoore:
1. Adds a new `revision` target that will initialize a new db migration file.
2. Adds help comments to the `revision` and `migrate` targets so they are listed when running `make help`


jeremywmoore:
```
% make
fix         Format the code
help        Print the help message
lint        Lint the code
migrate     Apply migrations
revision    Create a new migration
run         Run the app
test        Run the test suite
```

when running without specifying a message name:
```
% make revision
Makefile:31: *** 'message' must be set when invoking the revision target, eg `make revision message="short message"`.  Stop.
```

successful run:
```
% make revision message="short message"
. ./dev_env.sh && \
	poetry run flask db revision -m "short message"
  Generating /Users/jeremymoore/projects/hushline/migrations/versions/82c7cc8d5b80_short_message.py ...  done
```

rmlibre (changes requested):
Are there places in the docs where this should be added? I see this one:

https://github.com/scidsg/hushline/blob/95c6c5c76ab01212bb30e0220f5d3d11e6a60f1d/docs/DEV.md?plain=1#L51-L57

rmlibre (commented):
#### PS: 

I ran this & the migration commands I used for the last PR [#377], in the current [#411], but both failed to pick up any changes.

Do new tables need to be registered somewhere specific in order to be detected & integrated with the migrations?

docs/DEV.md
```diff
@@ -26,13 +26,6 @@ You also need Rust to install some of the Python dependencies. Install [rustup](
 curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
 ```
 
-Using pipx, install Python packages required for linting and static type checking:
```
jeremywmoore, line 29: these tools are installed by poetry

jeremywmoore:
Added the new target to the docs and then cleaned up the targets so users don't have to wrap the calls with `poetry run`.

jeremywmoore:
> #### PS:
> I ran this & the migration commands I used for the last PR [#377], in the current [#411], but both failed to pick up any changes.
> 
> Do new tables need to be registered somewhere specific in order to be detected & integrated with the migrations?

After creating the new revision file you should populate it with alembic operations to create the table and columns.

rmlibre (approved):
👍 I like the shortening of the commands

micahflee (approved):

PR #439, opened 2024-07-20
Update formatting for directory cards

glenn-sorrentino:
Better standardization of margin around visual elements.

### DIFF
--- a/hushline/static/css/style.css
+++ b/hushline/static/css/style.css
@@ -149,7 +149,7 @@ p.instr {
 .badgeContainer {
     display: flex;
     gap: .25rem;
-    margin-bottom: .75rem;
+    margin: .5rem 0;
     align-items: center;
 }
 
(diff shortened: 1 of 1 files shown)


glenn-sorrentino:
For whatever reason, the same logic to hide the badge container isn't working on the Verified tab of the directory, which in turn is causing weird formatting issues in the All tab. 

This PR temporarily removes the badge container logic in the directory.

hushline/static/css/style.css
```diff
@@ -149,7 +149,7 @@ p.instr {
 .badgeContainer {
     display: flex;
     gap: .25rem;
-    margin-bottom: .75rem;
+    margin: .5rem 0;
     align-items: center;
 }
```
ispykenny, line 155: Would also consider 

`.badgeContainer:empty {
    margin: 0;
}`

This pseudo selector applies specific styles if the element doesn't have any child elements.

glenn-sorrentino: Oh good call. Will add that JIC, but this actually made me reconsider the choice to only show verified badges on the all tab. Since this kind of a special designation, and the tab bar with the Verified label can be scrolled out of the viewport, maybe it's best to include them.

ispykenny (approved):

PR #445, opened 2024-07-22
Pagination implemented

ispykenny:
For testing
Update `usersPerPage` in directory.js
``` 
const usersPerPage = 50; // change this value for testing purposes
```

Tips: 
If there aren't any `verified` users in the response. We automatically click the `All` tab to show all list of users. 

Pagination links append `?page=index` to the url. the `index` value applies a `index * usersPerPage` offset in the user query

https://github.com/user-attachments/assets/69123f14-3713-4e8b-a915-8109eabc9285

### DIFF
--- a/hushline/routes.py
+++ b/hushline/routes.py
@@ -430,7 +430,7 @@ def init_app(app: Flask) -> None:
         return {"logged_in": logged_in}
 
     @app.route("/directory/users.json")
-    def directory_users() -> list[dict[str, str]]:
+    def directory_users() -> dict[str, int | list[dict[str, str]]]:
         count = request.args.get('count')
         offset = request.args.get('offset')
         users = get_directory_users(count=count, offset=offset)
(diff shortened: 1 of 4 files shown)


glenn-sorrentino (changes requested):
Looking great - couple things:

1. It looks like when clicking on a new page the tab briefly shows the verified tab then switches to all.
2. Also looks like the tabs alternate depending on the page?
Page 1 = All
Page 2 = All
Page 3 = Verified
Page 4 = All

glenn-sorrentino:
@ispykenny And now I read the PR about the alternating views if no records.

Instead, let's paginate per tab, and when getting to the end we can give an empty state message "You've reached the end of our verified users. Try the `All` tab for more. And the opposite for the other tab.

ispykenny:
> @ispykenny And now I read the PR about the alternating views if no records.
> 
> Instead, let's paginate per tab, and when getting to the end we can give an empty state message "You've reached the end of our verified users. Try the `All` tab for more. And the opposite for the other tab.

ahh yeah. I like this better. Will implement.

PR #450, opened 2024-07-23
test(2fa): improve stability with broader error code catches

rmlibre:
Sometimes flask returns error code 401 or 429 during 2FA tests, but the tests specified one or the other. The root cause of the inconsistency is unknown to me, but this PR should help prevent breaks to CI/CD tests & builds.

[Test results](https://github.com/rmlibre/hushline/actions/runs/10052255053/job/27783141070)

Cherry-picked from (e173071) on PR branch [#411]


jeremywmoore:
I don't know if this is the right approach as it is conflating rate limits with invalid tokens. Unexpected rate limits could mask actual test failures. Instead, I think something like https://github.com/scidsg/hushline/pull/460 might be a better approach.

PR #452, opened 2024-07-23
Display version string

micahflee:
This PR displays the version string in the footer of the template, like "Powered by Hush Line 0.1.10". The important change is this:

**Each time you make a release, you must define the release version in `hushline/version.py` before creating the git tag.**

What I recommend is for each release, we do the following:

- Create a release PR, like "Version 0.1.11" which includes the updated version string in `hushline/version.py`
- After the PR is merged, make a release tag

image


micahflee:
The dev deploy failed on the first commit but succeeded on the second commit. You can see it live here: https://hushline-dev-display-version-dnhhz.ondigitalocean.app/directory

jeremywmoore (approved):

PR #459, opened 2024-07-25
Pagination

ispykenny:
### We are now treating each tab as their our route. 

`Verified` tab  is the base `/directory` url
`All` tab is the all `/directory?viewing=all` url

Currently search loses it's value on page tab change which is expected as the entire page rerenders on page load. We could essentially move the value into `localStorage` and populate the field when it is changed. 

To test pagination with a different count other than `50` update the line below in `routes.py`

```
def usersPerPage() -> int:
        return 50
```


ispykenny:
@glenn-sorrentino I probably went down a rabbit hole with this but super curious on your thoughts! We could go back to having them be client-side-served but I think having the browser history and being able to set the active tab based on the url are nice haves.

glenn-sorrentino:
> @glenn-sorrentino I probably went down a rabbit hole with this but super curious on your thoughts! We could go back to having them be client-side-served but I think having the browser history and being able to set the active tab based on the url are nice haves.

Looking really good! I'm curious to hear @jeremywmoore and @micahflee's thoughts in relation to [this](https://github.com/scidsg/hushline/security/advisories/GHSA-7jmr-x386-c22m). 

One thing I notice is that when there are multiple pages of records and you search, when the input is cleared, the list shows all records, rather than the set from the page they're on. So in the movie below, pages have two records each, and when the query is cleared, four records are listed:

https://github.com/user-attachments/assets/10900b2a-7c9c-4e19-8d39-626735bacb5d

hushline/routes.py
```diff
@@ -417,16 +417,69 @@ def get_directory_users() -> list[User]:
...
+                .offset(offset)
+                .all()
+            )
+        return (
+            User.query.filter_by(show_in_directory=True, is_verified=True)
+            .order_by(User.is_admin.desc(), User.display_name.asc())
+            .limit(count)
+            .offset(offset)
+            .all()
+        )
```
jeremywmoore, line 461: Maybe something like this to simplify the parameter and avoid re-defining the query
```suggestion
            verified_only:bool = True
        ) -> list[User]:
        filter_by = {"show_in_directory": True}
        if verified_only:
            filter_by["verified_only"] = True
        return (
            User.query.filter_by(**filter_by)
            .order_by(User.is_admin.desc(), User.display_name.asc())
            .limit(count)
            .offset(offset)
            .all()
        )
```

hushline/routes.py
```diff
@@ -417,16 +417,69 @@ def get_directory_users() -> list[User]:
...
+
+    @app.route("/directory/pagination-users.json")
+    def pagination_users() -> list[dict[str, str]]:
+        count = request.args.get('count')
+        offset = request.args.get('offset')
```
jeremywmoore, line 466: Do these need defaults? Otherwise I think we may not get the intended value if the parameters aren't in the args.
```suggestion
        count = request.args.get('count', usersPerPage())
        offset = request.args.get('offset', 0)
```

jeremywmoore (commented):
Here's a review of just the python bits

PR #461, opened 2024-07-26
add hcaptcha to managed service

glenn-sorrentino:
- Added hcaptcha to the managed service.
- Vendored the JS 
- Tested and working
- Need to add secret to DO

hcaptcha


hushline/routes.py
```diff
@@ -120,15 +122,37 @@ def inbox() -> Response | str:
...
-    def submit_message(username: str) -> Response | str:
+    def submit_message(username: str) -> Union[Response, str]:
         form = MessageForm()
         user = User.query.filter_by(primary_username=username).first()
         if not user:
-            flash("🫥 User not found.")
+            flash("User not found.")
             return redirect(url_for("index"))
 
+        is_personal_server = os.getenv("IS_PERSONAL_SERVER", "false").lower() == "true"
```
micahflee, line 146: Same here. Instead of directly loading the environment, you can just do:

```py
is_personal_server = app.config["IS_PERSONAL_SERVER"]
```

micahflee: This is also nice because you don't need the like `"false").lower() == "true"` stuff each time

hushline/routes.py
```diff
@@ -120,15 +122,37 @@ def inbox() -> Response | str:
...
+        app.logger.debug(f"IS_PERSONAL_SERVER: {is_personal_server}")
+
         if form.validate_on_submit():
+            if not is_personal_server:
+                hcaptcha_response = request.form.get("h-captcha-response")
+                if not hcaptcha_response or not verify_hcaptcha(hcaptcha_response):
+                    flash("hCaptcha verification failed. Please try again.", "error")
```
micahflee, line 153: I'd just say: "CAPTCHA verification failed." No need to mention hCaptcha directly. (But I don't have strong opinions either way)

micahflee (changes requested):
I'm going to start with some feedback on the code itself, and then end with stepping back and considering the security issues that hCaptcha adds overall.

# The code

I've sprinkled some comments throughout the code to look at.

I think that if the `HCAPTCHA_SECRET_KEY` environment variable is not set, then we should disable hCaptcha. If it's set, we should use it. This way it can be an optional feature, and people can choose to not use hCaptcha when they're self-hosting, even if not using `IS_PERSONAL_SERVER`.

As it stands, there's no way for devs to test submitting messages without a valid `HCAPTCHA_SECRET_KEY`. If you try, even if you successfully submit the message, it says: "hCaptcha verification failed. Please try again."

If you do set a valid `HCAPTCHA_SECRET_KEY` though, it works as expected.

# Should we use a third-party CAPTCHA provider at all?

Even though you vendored the hCaptcha javascript, that code then loads even more javascript and makes various remote requests to hCaptcha's service. As long as you're using a third-party service, there's no way around this.

Just to see exactly what's happening, I opened Burp Suite and started intercepting all of my traffic to my dev server running this branch. The following screenshot shows all of the http requests my browser made as I filled out a CAPTCHA and submitted a message.

image

As you can see it, it loads `https://newassets.hcaptcha.com//captcha/v1/988e468/hcaptcha.js` along with multiple other javascript source files, and it executes them in the browser.

This basically puts hCaptcha in a super privileged position. If they choose to, they could respond with malicious javascript code that can do anything the current user can do. If they're logged into their own Hush Line account while the javascript is running, it can for example read all their messages.

It's possible that the javascript getting loaded is in an iframe, which provides some protection (the inner frame shouldn't be able to run arbitrary code in the parent frame), but most likely there are still some ways that hCaptcha could still do evil stuff if they chose too.

Hush Line's threat model isn't as strict as SecureDrop's and we're putting trust in third-party hosting providers, but I do think it's prudent to avoid third-party javascript at least. In practice, I think this means instead of using a service like hCaptcha, ReCAPTCHA, or anything else, it's a better option to build our own local CAPTCHA service directly into Hush Line. There may be some open source options we can use too. These likely won't be as accurate as the third-party services, but it's how we can prevent third-parties from having any access.

glenn-sorrentino:
I'm okay holding on this. Thanks for the thorough review!

PR #465, opened 2024-07-28
Dark Mode

glenn-sorrentino:
Added dark mode:

Frame 298


hushline/static/css/style.css
```diff
@@ -1,12 +1,24 @@
 :root {
```
ispykenny, line 1: This all works as expected.

My approach on handling theming is:

For example: 
instead of `--color-text-dark` and `--color-text-light` it's just `--color-primary-text`

and then all you do is set the 
```
@media (prefers-color-scheme: light) {
  :root {
      --color-primary-text: black;
   }
}

@media (prefers-color-scheme: dark) {
  :root {
      --color-primary-text: white;
   }
}

.selector {
   color: var(--color-primary-text);
}
```

that way you can avoid duplicate css for each color scheme and just update only the `--var` values within the `@media` query.


Dark theme looks good! Tested and looks clean.

ispykenny (approved):

hushline/static/css/style.css
glenn-sorrentino: Great recommendation. Will clean up the variables in a subsequent PR.

PR #468, opened 2024-07-31
fix: ensure new passwords are distinct

rmlibre:
This PR ensures that the change password flow is consistent, safely & consistently flashes relevant error messages, and doesn't allow for a change to an identical password.

#### Passing Workflows:

- [x] [Run Linters & Tests](https://github.com/rmlibre/hushline/actions/runs/10187133104/job/28180490773) on (https://github.com/scidsg/hushline/pull/468/commits/9dfcdc5e7c84a8467b7d9260f004c8680098f3b2)

### DIFF
--- a/hushline/forms.py
+++ b/hushline/forms.py
@@ -1,4 +1,6 @@
 import re
+from hmac import compare_digest as bytes_are_equal
+from typing import Callable
 
 from flask_wtf import FlaskForm
 from wtforms import Field, Form, StringField
@@ -23,6 +25,16 @@ class ComplexPassword:
             and re.search("[^A-Za-z0-9]", password)
         ):
             raise ValidationError(self.message)
+
+
+def is_valid_password_swap(
+    *, check_password: Callable[[str], bool], old_password: str, new_password: str
+) -> bool:
+    # since the passwords can be of different lengths, the equality test must occur *iif*
+    # the correctness test passes, since timing differences leak length information
+    if check_password(old_password):
+        return not bytes_are_equal(old_password.encode(), new_password.encode())
+    return False
 
 
 class TwoFactorForm(FlaskForm):
--- a/hushline/forms.py
+++ b/hushline/forms.py
@@ -7,6 +7,10 @@ from wtforms import Field, Form, StringField
 from wtforms.validators import DataRequired, Length, ValidationError
 
 
+class WrongPassword(ValueError):
+    pass
+
+
 class ComplexPassword:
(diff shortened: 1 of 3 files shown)


rmlibre (commented):
Cleanup question.

hushline/settings.py
```diff
@@ -176,12 +176,20 @@ def index() -> str | Response:  # noqa: PLR0911, PLR0912
 
             # Handle Change Password Form Submission
```
rmlibre, line 177: Is this section needed if the ``change_password`` route function exists? I tried removing this section to see if the tests would pass, and they do. I'm not sure if that's because this isn't needed, or just because it isn't tested.

rmlibre (commented):
Cleanup question

hushline/settings.py
jeremywmoore: Agreed, it looks like only changes to display name, username, directory visibility, and bio are handled via the `/settings` route, while 2fa, password changes, smtp settings, pgp settings, account deletion, verified user and admin user settings are handled by dedicated routes.

hushline/forms.py
```diff
@@ -25,5 +31,15 @@ def __call__(self, form: Form, field: Field) -> None:
             raise ValidationError(self.message)
 
 
+def is_valid_password_swap(
+    *, check_password: Callable[[str], bool], old_password: str, new_password: str
```
jeremywmoore, line 35: I don't love passing a callable here, is the goal to keep all password validation logic in one part of the code?

hushline/forms.py
```diff
@@ -25,5 +31,15 @@ def __call__(self, form: Form, field: Field) -> None:
...
+) -> bool:
+    # since the passwords can be of different lengths, the equality test must occur *iif*
+    # the correctness test passes, since timing differences leak length information
+    if check_password(old_password):
+        return not bytes_are_equal(old_password.encode(), new_password.encode())
+    raise WrongPassword
```
jeremywmoore, line 41: Raising an error isn't really consistent with the existing code style. I am open to refactoring the flow control to be error-based, but I'd prefer that as a separate PR.

hushline/forms.py
rmlibre: Yeah, that was the idea. New refactorings also made it unnecessary. I removed this pattern in the new PR #635

hushline/forms.py
rmlibre: I also removed this pattern in the new PR #635

PR #471, opened 2024-08-01
Adopt sql alchemy session api

jeremywmoore:
The PR refactors the hushline codebase onto sqlalchemy's new session API. This interface ends up looking much more like the sql that it generates and should reduce the "magic" of the query builder. Along with this, I've updated the models to use the new declarative definition, with the main result being improved type-checking.

At the same time, this refactors the deprecated use of `datetime.utcnow()` in favor of `datetime.now(datetime.UTC)`

 In the process of refactoring, a few issues were highlighted related to 2fa validation if it doesn't get setup correctly.

### DIFF
--- a/hushline/__init__.py
+++ b/hushline/__init__.py
@@ -54,7 +54,7 @@ def create_app() -> Flask:
     @app.context_processor
     def inject_user() -> dict[str, Any]:
         if "user_id" in session:
-            user = User.query.get(session["user_id"])
+            user = db.session.get(User, session["user_id"])
             return {"user": user}
         return {}
 
--- a/hushline/routes.py
+++ b/hushline/routes.py
@@ -104,7 +104,7 @@ def init_app(app: Flask) -> None:
             messages = db.session.scalars(
                 db.select(Message).filter_by(user_id=logged_in_user.id).order_by(Message.id.desc())
             ).all()
-            secondary_users_dict = {su.id: su for su in list(logged_in_user.secondary_usernames)}
+            secondary_users_dict = {su.id: su for su in logged_in_user.secondary_usernames}
 
         return render_template(
             "inbox.html",
@@ -453,7 +453,7 @@ def init_app(app: Flask) -> None:
         return {"logged_in": logged_in}
 
     @app.route("/directory/users.json")
-    def directory_users() -> list[dict[str, str]]:
+    def directory_users() -> list[dict[str, str | bool | None]]:
         return [
             {
                 "primary_username": user.primary_username,
--- a/hushline/routes.py
+++ b/hushline/routes.py
@@ -16,7 +16,6 @@ from flask import (
     url_for,
 )
 from flask_wtf import FlaskForm
-from sqlalchemy import func, select
 from werkzeug.wrappers.response import Response
 from wtforms import Field, Form, PasswordField, StringField, TextAreaField
 from wtforms.validators import DataRequired, Length, Optional, ValidationError
@@ -103,7 +102,7 @@ def init_app(app: Flask) -> None:
 
         if logged_in_user:
             messages = db.session.scalars(
-                select(Message).filter_by(user_id=logged_in_user.id).order_by(Message.id.desc())
+                db.select(Message).filter_by(user_id=logged_in_user.id).order_by(Message.id.desc())
             ).all()
             secondary_users_dict = {su.id: su for su in list(logged_in_user.secondary_usernames)}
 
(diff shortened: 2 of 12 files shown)


hushline/__init__.py
```diff
@@ -54,7 +54,7 @@ def page_not_found(e: Exception) -> Response:
     @app.context_processor
     def inject_user() -> dict[str, Any]:
         if "user_id" in session:
-            user = User.query.get(session["user_id"])
+            user = db.session.get(User, session["user_id"])
```
jeremywmoore, line 57: https://docs.sqlalchemy.org/en/20/orm/session_api.html#sqlalchemy.orm.Session.get

hushline/routes.py
```diff
@@ -96,34 +96,30 @@ def inbox() -> Response | str:
...
             return redirect(url_for("inbox"))
 
-        primary_user = User.query.get(logged_in_user_id)
-        if primary_user:
-            messages = (
-                Message.query.filter_by(user_id=primary_user.id).order_by(Message.id.desc()).all()
-            )
-            secondary_users_dict = {su.id: su for su in primary_user.secondary_usernames}
+        if logged_in_user:
+            messages = db.session.scalars(
```
jeremywmoore, line 104: `session.scalars()` is shorthand for `session.execute().scalars()` which executes the statement and maps the rows back to the resulting objects.

hushline/routes.py
```diff
@@ -96,34 +96,30 @@ def inbox() -> Response | str:
...
+                db.select(Message).filter_by(user_id=logged_in_user.id).order_by(Message.id.desc())
+            ).all()
+            secondary_users_dict = {su.id: su for su in list(logged_in_user.secondary_usernames)}
```
jeremywmoore, line 107: This line needs extra eyes on it. MyPy didn't like iterating over the `secondary_usernames` type, I _think_ create a list  from it should resolve the lazy references.

hushline/routes.py
```diff
@@ -231,8 +227,12 @@ def register() -> Response | str | tuple[Response | str, int]:
 
             invite_code_input = form.invite_code.data if require_invite_code else None
             if invite_code_input:
-                invite_code = InviteCode.query.filter_by(code=invite_code_input).first()
-                if not invite_code or invite_code.expiration_date < datetime.utcnow():
+                invite_code = db.session.scalars(
+                    db.select(InviteCode).filter_by(code=invite_code_input).limit(1)
+                ).first()
+                if not invite_code or invite_code.expiration_date.replace(
```
jeremywmoore, line 233: the `datetime.replace()` call here is to convert the naive datetime to one with a timezone. We could update the schema to return non-naive datetimes, but that seemed worth deferring for now.

hushline/routes.py
```diff
@@ -305,13 +309,13 @@ def login() -> Response | str:
         )
 
     @app.route("/verify-2fa-login", methods=["GET", "POST"])
-    def verify_2fa_login() -> Response | str | tuple[Response | str, int]:
+    def verify_2fa_login() -> Response | str | tuple[Response | str, int]:  # noqa: PLR0911
```
jeremywmoore, line 312: can't remember why this was failing to lint

hushline/routes.py
```diff
@@ -344,12 +352,15 @@ def verify_2fa_login() -> Response | str | tuple[Response | str, int]:
...
             # If there were 5 failed logins in the last 30 seconds, don't allow another one
-            failed_logins = (
-                AuthenticationLog.query.filter_by(user_id=user.id, successful=False)
-                .filter(AuthenticationLog.timestamp > datetime.now() - timedelta(seconds=30))
-                .count()
+            failed_logins = db.session.scalar(
+                db.select(
+                    db.func.count(AuthenticationLog.id)
+                    .filter(AuthenticationLog.user_id == user.id)
+                    .filter(AuthenticationLog.successful == db.false())
```
jeremywmoore, line 359: this one gave me a bit of trouble because if you use `False` instead of `db.false()` the linter is unhappy and recommends something that isn't equivalent.

hushline/routes.py
```diff
@@ -344,12 +352,15 @@ def verify_2fa_login() -> Response | str | tuple[Response | str, int]:
...
+                    .filter(AuthenticationLog.timestamp > datetime.now() - timedelta(seconds=30))
+                )
             )
-            if failed_logins >= 5:  # noqa: PLR2004
+            if failed_logins is not None and failed_logins >= 5:  # noqa: PLR2004
```
jeremywmoore, line 363: can't remember why the linter didn't like this

hushline/routes.py
jeremywmoore: ah, it's "too may return statements"

hushline/routes.py
jeremywmoore: This is no longer an issue after adopting the new declarative table definition

micahflee (approved):

PR #485, opened 2024-08-13
Accessible Math CAPTCHA

glenn-sorrentino:
Adds CAPTCHA which serves a simple math problem through an accessible input.

math


glenn-sorrentino:
Updated with better voice-over experience

https://github.com/user-attachments/assets/d74ccb8e-331e-4a6d-84fd-159cd3d89467

glenn-sorrentino:
Updated with better voice-over experience

No audio, but you can see the caption...

https://github.com/user-attachments/assets/d2820f8b-9bd2-4ada-bac4-a295fd82b6c3

micahflee (approved):
Obviously this captcha is trivially defeated if someone just writes some code to defeat it, however my guess is this will still prevent a lot of automated spam. We can always improve on it later, and I think this is good to merge.

jeremywmoore (approved):

PR #488, opened 2024-08-15
Implement Server-Side Session Management with Redis and Improve Security for Session Cookies

glenn-sorrentino:
1.	**Implement Server-Side Session Management with Redis:**
- Switched from client-side session cookies to server-side session management using Redis.
- Configured Flask to store session data in Redis, ensuring that sensitive session information is securely managed server-side.
2.	**Enhanced Cookie Security:**
- Ensured session cookies are HTTP-only, secure, and have the SameSite=Lax attribute.
- The session cookie now only contains a reference to the server-side session, minimizing exposure of sensitive data.
3.	**Refactoring of Session Data:**
- Removed sensitive information like is_admin from being stored in the session cookie.
- Refactored the application to fetch critical flags like is_admin directly from the database when needed.
4.	**Environment Configuration:**
- Added Redis server connection details to the environment configuration, ensuring that Redis is correctly set up and running.

## Testing

- Verified that session cookies are now server-side managed and do not expose sensitive data.
- Inspected the session cookie to ensure that it is secure and does not contain plaintext sensitive information.

**Before:**
Screenshot 2024-08-14 at 5 58 39 PM

**After:**
Screenshot 2024-08-14 at 5 59 56 PM

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -10,6 +10,18 @@ on:
 jobs:
  run-lint-and-tests:
     runs-on: ubuntu-latest
+
+    services:
+      redis:
+        image: redis:latest
+        ports:
+          - 6379:6379
+        options: >-
+          --health-cmd "redis-cli ping"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+          
     steps:
       - uses: actions/setup-python@v4
         with:
--- a/dev_env.sh
+++ b/dev_env.sh
@@ -4,3 +4,4 @@ export SECRET_KEY=cb3f4afde364bfb3956b97ca22ef4d2b593d9d980a4330686267cabcd2c0be
 export SQLALCHEMY_DATABASE_URI=sqlite:///hushline.db
 export REGISTRATION_CODES_REQUIRED=False
 export SESSION_COOKIE_NAME=session
+export REDIS_URL=redis://localhost:6379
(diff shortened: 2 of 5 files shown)


jeremywmoore:
Is this change seen as a pre-requisite for captcha support? Redis adds a significant amount of complexity to the site and a highly available pair of instances costs $60/mo on digital ocean. Before we go this route, I'd consider encrypting sensitive fields within the session cookie.

micahflee (changes requested):
We could have way lower-availability instances of Redis (just using docker an App Platform) for $6/month for dev, but we'd want something more robust for prod. I agree that adding Redis adds a lot of complexity (and will also break the personal server).

However, Redis isn't the only option for a server-side session store. It looks like sqlalchemy is an option, and something we're already using: https://flask-session.readthedocs.io/en/latest/installation.html#available-storage-options

@glenn-sorrentino do you want to refactor this to store session data in the existing sql database?

micahflee:
Also, #490 already removes everything from the session except `user_id`, `is_authenticated`, and `username`. (We can also remove `username` but it would be more complicated since the base template relies on it.)

glenn-sorrentino:
> We could have way lower-availability instances of Redis (just using docker an App Platform) for $6/month for dev, but we'd want something more robust for prod. I agree that adding Redis adds a lot of complexity (and will also break the personal server).
> 
> However, Redis isn't the only option for a server-side session store. It looks like sqlalchemy is an option, and something we're already using: https://flask-session.readthedocs.io/en/latest/installation.html#available-storage-options
> 
> @glenn-sorrentino do you want to refactor this to store session data in the existing sql database?

Do you wanna tag in? Guessing you'll be 10x faster at this than me :)

PR #490, opened 2024-08-16
More strictly enforce authentication, and clean up routes

micahflee:
While implementing an unrelated feature I discovered a bunch of problems and inconsistencies with the routes, as they're defined. This PR tries to fix as many of them as possible. It does the following:

# Updated authentication

Many of the routes included the confusingly-named `@require_2fa` decorator, even routes that were unrelated to 2FA. What this route actually did was ensure that the user was logged in, with extra logic to ensure that 2FA was verified if it's enabled. This was all too complicated.

I deleted the `@require_2fa` decorator and made two new ones, `@authentication_required` and `@admin_authentication_required`. If `user_id` is set in the session and `is_authenticated` in the session is true, then the user is logged in. I moved all of the 2FA logic into the login flow itself.

Before, during login the code set the values `2fa_required`, `2fa_verified`, and `is_admin` to the session. This PR removes these from the session. Instead it works like this:

- When you start the login process, is sets `user_id` to your user ID. If you have 2FA, it sets `is_authenticated` to false and redirects to the `verify_2fa_login` route, and otherwise it sets `is_authenticated` to true.
- From the `verify_2fa_login` route, if you verify 2FA, it sets `is_authenticated` to true.
- The one bit of code that used `is_admin` now figures out if the user is admin by checking the database and not the session.

# Required authentication everywhere it should be required

This adds the `@authentication_required` decorator to a bunch of routes that didn't require authentication before, when they should have, including:

- `/verify-2fa-login`
- `/settings/update_directory_visibility`
- `/settings/toggle-2fa`
- `/settings/confirm_disable_2fa`
- `/settings/verify-2fa-setup`

These routes mostly seemed to expect `user_id` to be in the session to work, but it's entirely possible there are code paths that could be exploited. For example, here I am not logged into an account at all, but looking at the disable 2FA page:

image

# Fixed the `/inbox` route to not take the username as client-side input

Before, `/inbox` required a `username` URL parameter (like http://localhost:8080/inbox?username=user), and if you try loading it without it (like http://localhost:8080/inbox), the browser gets stuck an in infinite redirect loop:

image

There's no need to ask the user to provider the username of what inbox they're looking at, and then confirm that it matches what's in the database. It's much simpler to just pull this data from the database directly.

# Deleted completely unused routes

I deleted these routes because they weren't being used, and the code was duplicated in different places:

- `/show-qr-code`, because the QR code logic is actually handled in `/enable-2fa`
- `/settings/change-username`, because this POST request is handled in `/settings` instead

I don't know how these ended up there, but it's good to delete unused code as it could have bugs that could some day get exploited.

# Small fixes

- Prevents logged-in users from registering an account (`/register`, GET and POST).
- Prevents logged-in users from trying to login again (`/login`, GET and POST).

### DIFF
--- a/hushline/settings.py
+++ b/hushline/settings.py
@@ -377,7 +377,7 @@ def create_blueprint() -> Blueprint:
         session.pop("is_setting_up_2fa", None)
         return redirect(url_for("logout"))
 
-    @bp.route("/update_pgp_key", methods=["GET", "POST"])
+    @bp.route("/update-pgp-key", methods=["POST"])
     @authentication_required
     def update_pgp_key() -> Response | str:
         user_id = session.get("user_id")
@@ -406,9 +406,10 @@ def create_blueprint() -> Blueprint:
             db.session.commit()
             flash("👍 PGP key updated successfully.")
             return redirect(url_for(".index"))
-        return render_template("settings.html", form=form)
 
-    @bp.route("/update_smtp_settings", methods=["GET", "POST"])
+        return redirect(url_for(".index"))
+
+    @bp.route("/update-smtp-settings", methods=["POST"])
     @authentication_required
     def update_smtp_settings() -> Response | str:
         user_id = session.get("user_id")
@@ -421,10 +422,7 @@ def create_blueprint() -> Blueprint:
             return redirect(url_for(".index"))
 
         # Initialize forms
-        change_password_form = ChangePasswordForm()
-        change_username_form = ChangeUsernameForm()
         smtp_settings_form = SMTPSettingsForm()
-        pgp_key_form = PGPKeyForm()
 
         # Handling SMTP settings form submission
         if smtp_settings_form.validate_on_submit():
@@ -439,23 +437,7 @@ def create_blueprint() -> Blueprint:
             flash("👍 SMTP settings updated successfully")
             return redirect(url_for(".index"))
 
-        # Prepopulate SMTP settings form fields
-        smtp_settings_form.email.data = user.email
-        smtp_settings_form.smtp_server.data = user.smtp_server
-        smtp_settings_form.smtp_port.data = user.smtp_port
-        smtp_settings_form.smtp_username.data = user.smtp_username
-        # Note: Password fields are typically not prepopulated for security reasons
-
-        pgp_key_form.pgp_key.data = user.pgp_key
-
-        return render_template(
-            "settings.html",
-            user=user,
-            smtp_settings_form=smtp_settings_form,
-            change_password_form=change_password_form,
-            change_username_form=change_username_form,
-            pgp_key_form=pgp_key_form,
-        )
+        return redirect(url_for(".index"))
 
     @bp.route("/delete-account", methods=["POST"])
     @authentication_required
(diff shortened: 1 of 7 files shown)


jeremywmoore:
There's a strange state a user can get in here if they remember their login, but can't verify 2fa. In this state the user cannot logout but also cannot register an account as they are partially logged in.

hushline/utils.py
```diff
@@ -8,16 +8,32 @@
...
+
+        return f(*args, **kwargs)
+
+    return decorated_function
+
+
+def admin_authentication_required(f: Callable[..., Any]) -> Callable[..., Any]:
+    @wraps(f)
+    def decorated_function(*args: Any, **kwargs: Any) -> Any:
+        if "user_id" not in session or not session.get("is_authenticated", False):
```
jeremywmoore, line 29: Do we want to check for `user_id in session and not session.get("is_authenticated")` and always redirect to the `verify-2fa` route in that case?

hushline/utils.py
```diff
@@ -8,16 +8,32 @@
...
+            return redirect(url_for("login"))
+
+        user = db.session.query(User).get(session["user_id"])
+        if not user or not user.is_admin:
+            flash("Unauthorized access.", "error")
```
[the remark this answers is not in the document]
micahflee: Yes good idea

hushline/utils.py
micahflee: Good idea

micahflee:
@jeremywmoore I've fixed all of the issues that you mentioned:

- I added a cancel link to the `/verify-2fa-login` page that logs you out
- I made `@admin_authentication_required` return 403 on failure, and also use `@authentication_required`
- I made `@authentication_required` redirect to `/verify-2fa-login` on partial-login but not authenticated

micahflee:
I've thrown in one more fix into this PR. The `update_pgp_key` and `update_smtp_settings` routes were accepting both GET and POST methods, but GET caused a crash and was a code path that was never used. This fixes that.

jeremywmoore:
Hmm, clicking the cancel link on the verify-2fa page redirects back to the verify-2fa page. Also, could we apply the button style to the link?

hushline/settings.py
```diff
@@ -406,9 +406,10 @@ def update_pgp_key() -> Response | str:
             db.session.commit()
             flash("👍 PGP key updated successfully.")
             return redirect(url_for(".index"))
-        return render_template("settings.html", form=form)
 
-    @bp.route("/update_smtp_settings", methods=["GET", "POST"])
+        return redirect(url_for(".index"))
+
+    @bp.route("/update-smtp-settings", methods=["POST"])
```
jeremywmoore, line 412: do we need to update the routes in the `settings.html` template to reflect these changes?

jeremywmoore (changes requested):

hushline/settings.py
jeremywmoore: after reading the docs, it looks like ~url_for` accepts the name of the function, not the url so we don't need to make any changes to the template.

jeremywmoore (approved):

PR #491, opened 2024-08-17
Proton Mail key discovery

### LINKED ISSUE #486: Easy Proton Mail key discovery
PGP sucks. Configuring Hush Line to forward encrypted messages to Proton Mail, using the user's Proton-managed PGP key, makes it a lot easier. But it's a manual process for Proton users to figure out how to find their public key, export it, and copy and paste it into Hush Line. This should be automatic.

In the message encryption settings, there should be an option to use your Proton Mail PGP key. When you enter an email address hosted on Proton, it should use Proton's key server (just GET requests like `https://mail-api.proton.me/pks/lookup?op=get&search=user.email@proton.me`) to automatically get the email address's public key, if it's available.

micahflee:
Fixes #486.

This PR is based on #490, so that one should be reviewed and merged first before this one.

In the Settings, Email & Encryption tab, this now adds subtabs (and some css for subtabs). Under Message Encryption, there's now subtabs for Proton Mail and PGP Public Key.

If no PGP key is set, it defaults to Proton Mail. If a public key is set, it defaults to PGP Public Key.

From the Proton Mail subtab, you can search for any email address hosted on Proton Mail and it will use Proton's PGP key server to look up the public key and automatically add it. If you search for an email address that isn't hosted by Proton, it will give an error.

Demo:

[Screencast from 2024-08-16 17-27-23.webm](https://github.com/user-attachments/assets/fe0c0d1f-2ad6-4654-8e8e-a2a584a3f984)


glenn-sorrentino (changes requested):
I have a few interface and interaction updates in #492

hushline/settings.py
```diff
@@ -49,6 +50,10 @@ class SMTPSettingsForm(FlaskForm):
     smtp_password = PasswordField("SMTP Password", validators=[DataRequired()])
 
 
+class PGPProtonForm(FlaskForm):
+    email = StringField("Proton Email Address", validators=[DataRequired()])
```
jeremywmoore, line 54: Should we use the [Email validator](https://wtforms.readthedocs.io/en/3.0.x/validators/#wtforms.validators.Email) supported by WTForms?

hushline/settings.py
```diff
@@ -457,8 +384,53 @@ def verify_2fa_setup() -> Response | str:
...
+    @authentication_required
+    def update_pgp_key_proton() -> Response | str:  # noqa: PLR0911
+        user_id = session.get("user_id")
+        if not user_id:
+            flash("⛔️ User not authenticated.")
+            return redirect(url_for("login"))
+
+        user = db.session.get(User, user_id)
+        form = PGPProtonForm()
+        if form.validate_on_submit():
```
jeremywmoore, line 397: can we invert this check and return immediately if validation fails? I find it much easier to follow the flow when we check for errors immediately as we can return early from the conditional and then continue with normal execution without increasing the nesting depth.

hushline/settings.py
```diff
@@ -457,8 +384,53 @@ def verify_2fa_setup() -> Response | str:
...
+            email = form.email.data
+
+            if is_valid_email(email):
```
jeremywmoore, line 400: Let's move this into a form validator?

hushline/settings.py
```diff
@@ -457,8 +384,53 @@ def verify_2fa_setup() -> Response | str:
...
+                        f"https://mail-api.proton.me/pks/lookup?op=get&search={email}", timeout=5
+                    )
+                except requests.exceptions.RequestException as e:
+                    current_app.logger.error(f"Error fetching PGP key from Proton Mail: {e}")
+                    flash("⛔️ Error fetching PGP key from Proton Mail.")
+                    return redirect(url_for(".index"))
+                if r.status_code == 200:  # noqa: PLR2004
+                    pgp_key = r.text
+                    if is_valid_pgp_key(pgp_key):
+                        if user:
```
jeremywmoore, line 413: let's move this check up to immediately after the db query? If there is no matching user, should we clear the session and redirect to the login page?

hushline/settings.py
```diff
@@ -457,8 +384,53 @@ def verify_2fa_setup() -> Response | str:
...
+                            user.pgp_key = pgp_key
+                    else:
+                        flash("⛔️ No PGP key found for the email address.")
+                        return redirect(url_for(".index"))
+                else:
+                    flash("⛔️ This isn't a Proton Mail email address.")
+                    return redirect(url_for(".index"))
+            else:
+                # If the PGP key is invalid
```
jeremywmoore, line 422: This comment doesn't match the error message

hushline/settings.py
micahflee: Resolved in https://github.com/scidsg/hushline/pull/491/commits/8931a8aa54f6c532af7665696e789f7263e19573

hushline/settings.py
micahflee: Resolved in https://github.com/scidsg/hushline/pull/491/commits/8931a8aa54f6c532af7665696e789f7263e19573

hushline/settings.py
micahflee: Resolved in https://github.com/scidsg/hushline/pull/491/commits/9f423c5f200968fc6d08b0b4fbbbd28b47019009

hushline/settings.py
micahflee: I think Glenn fixed this one in his UI PR...

glenn-sorrentino (approved):

jeremywmoore (approved):

PR #494, opened 2024-08-17
Update SMTP UI

glenn-sorrentino:
- Moves "Update" button into parent `fieldset`. 
- Updates `select` menu for more consistent styling
- Updates toggle ui order to match directory opt-in layout

smtp

### DIFF
--- a/hushline/static/css/style.css
+++ b/hushline/static/css/style.css
@@ -102,8 +102,10 @@
     }
 
     input,
-    textarea {
+    textarea,
+    select {
         border: var(--border);
+        background-color: white;
     }
 
     code,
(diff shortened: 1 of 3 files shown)


jeremywmoore:
If the update button is hidden in the fieldset the ux becomes a bit more complicated to disable email forwarding. Does toggling forwarding off automatically submit the form? Toggling on can't as it requires the user to enter their email before submitting.

glenn-sorrentino:
Yeah toggling off should automatically submit. Typically any toggle doesn’t require a submit. The directory opt-in has the ux to shoot for. 
 On Sat, Aug 17, 2024 at 4:16 AM, Jeremy Moore ***@***.***&gt; wrote:  If the update button is hidden in the fieldset the ux becomes a bit more complicated to disable email forwarding. Does toggling forwarding off automatically submit the form? Toggling on can't as it requires the user to enter their email before submitting.

glenn-sorrentino:
@jeremywmoore The latest updates to `settings.js` submits the form when toggling off.

jeremywmoore (approved):

PR #499, opened 2024-08-20
Add Tor Banner

glenn-sorrentino:
Includes a global banner recommending that the user download and use Tor for increased anonymity.

@ispykenny is there anything we need to do from an a11y perspective?

Screenshot 2024-08-20 at 10 52 09 AM


ispykenny:
Looks good to me!

ispykenny (approved):

PR #510, opened 2024-08-22
Add Proton to docs

glenn-sorrentino:
Adds proton to docs.


docs/1-getting-started.md
```diff
@@ -4,7 +4,9 @@ Hush Line is a free and open-source, whistleblowing platform for organizations o
 
 ## Contents
 
-1. [Email Encryption](#email-encryption)
+1. Email Encryption
+   - [Option 1: Proton _(Recommended)_](#option-1-proton)
+   - [Option 2: Mailvelope and Gmail](#option-2-mailvelope--gmail)
```
jeremywmoore, line 9: Should we mention gmail here? It doesn't seem like any of the instructions actually involve gmail specifically (barring the example smtp settings). Maybe "webmail" is a better option since that's more generally what mailvelope supports? I also wonder if we want to point folks to something like [GPGSuite](https://gpgtools.org/) which includes an extension for Apple's Mail client to support PGP decryption.

glenn-sorrentino: Ah - good catch - Proton bundles forwarding in, and Mailvelope doesn't - I'll update the structure for more consistency

micahflee (changes requested):
I actually think this is a good time to totally update the getting started docs. I've been reading the changes from your branch here: https://github.com/scidsg/hushline/blob/glenn-sorrentino-patch-3/docs/1-getting-started.md

Here are a few more things I think you should change.

In the header of the "Email Encryption" section, right before "Option 1: Proton", it says: "We'll use Mailvelope, an open-source, cross-browser extension that allows you to decrypt messages directly in the app." That should be moved down to the Mailvelope section.

In "Option 2: Mailvelope and Gmail" should probably be "Mailvelope" or "Mailvelope and Webmail", as Mailvelope will work with any webmail provider.

I think you can actually remove some of the Firefox-only Mailvelope instructions and instead link to https://mailvelope.com/en/help, which should stay up-to-date and also includes instructions on how to set it up for various email providers. 

The screenshots for adding your key to Hush Line in Step 4 are out-of-date.

The Proton section includes enabling Email Forwarding, but the Mailvelope section doesn't. I think it makes sense to add this to Mailvelope, and then show an example of receiving encryption messages via webmail and Mailvelope too.

Then finally, I think you can just remove the Message Forwarding section because this isn't necessary anymore on the hosted service. You could maybe add documentation about custom SMTP for advanced users, but it's pretty self-explanatory so maybe it's not necessary.

docs/1-getting-started.md
micahflee: The most popular PGP mail client is Thunderbird, so you could include instructions for that too potentially

micahflee (changes requested):
A few small changes.

In #511 I changed the nav link "Submit Message" to be "Profile". In option 1 step 4, it says to click "Submit Message", so we might as well take this change to update it to say to click "Profile" and to update the screenshots.

In option 2 step 3, "Add your key to Hush Line", the text explains how to add your key but the screenshots are for actually submitting a message. I think it makes sense to add screenshots for how to add your key, then add a step for submitted the test message.

Also, I find that showing two screenshots side-by-side makes the details quite difficult to see. I think it might be better to stack them vertically, just so you can easily ready the text inside the screenshots.

And while not strictly necessary, I think it might be good to add one more optional step in the Mailvelope section for forwarding email, with just an example for how you can get encrypted tips sent to Gmail, and you can use Mailvelope to decrypt them.

glenn-sorrentino:
Reopening with a cleaner branch

PR #511, opened 2024-08-22
Require PGP

### LINKED ISSUE #487: Require PGP on hosted service
It's optional for users to use PGP. This is fine for self-hosted contexts, especially the personal server where the server is literally a Raspberry Pi at your house. But this is not good for the tips.hushline.app hosted service, as it means users will have plaintext messages from sources hosted in the cloud, which are vulnerable to data requests.

We should add a new `REQUIRE_PGP` environment variable, and enable it on the hosted service. This should make it so users can have accounts, but they can't accept any message submissions until they include a PGP key. Users can still be listed in the directory, however.

micahflee:
Resolves #487.

This adds a `REQUIRE_PGP` environment variable. If it's enabled, then it makes it so that people can only submit messages to users with PGP keys, though users without PGP keys still have profiles. It also adds 4 optional extra profile fields (along with a database migration). It also updates the route from `/submit_message/{username}` to `/profile/{username}`.

Here are all of the changes in more details:

In the directory, instead of "Send a Message" links under each user, it changed those links to "View Profile".

image

When you click that link, it goes to `/profile/{username}` instead of `/submit_message/{username}`:

image

**Open question: Are we okay with changing this route? This means if anyone has their current Hush Line profile links anywhere, they will be broken links. This probably doesn't matter since we have so few users, and a change like this would be a bigger deal later on. What do you think? Should make `/submit_message/*` redirect to `/profile/*`?** 

Now, if I set `REQUIRE_PGP=true` and restart the server, this is what viewing my profile looks like:

image

Specifically, if a user doesn't have a PGP key, it doesn't include the submit message form (or the javascript that powers it), and it says:

> You can't send encrypted messages to this user through Hush Line. Find secure contact information for them in their profile above.

There's also server-side validation so that if someone tries making a POST request to submit a message anyway it's rejected.

If you're logged into your account and look at your own profile, the "Only visible to you" text has been updated:

image

It says:

> **Only visible to you:** In order to protect your sources, you need to add a PGP key to your account before anyone can send you tips. Without a PGP key, your profile can still be listed in the Hush Line directory. Make sure to include a secure way to contact you, such as a Signal username, in your profile.

If you go to Settings > Profile, there are now extra fields:

image

You can choose whatever labels and values you want, like this for example:

image

When someone views your profile, if you have any extra fields set, it displays them:

image

It only displays a field if both a label and value are set.

If the value starts with `https://` then it makes it a link.

**Are there any security issues we should consider about letting users add arbitrary links? I'm only checking for `https://`, not other protocols (like `http://`), and if you do add a link it includes the attibutes `target="_blank" rel="noopener noreferrer"`.**

It also escapes the content so it shouldn't be vulnerable to XSS. If you try putting javascript in these fields it will just display the javascript instead of executing it.

Now, if you add a PGP key, it displays the submit form again:

image

And if you view your own profile while logged in, the "Only visible to you" text changes:

image

It says what it originally said:

> **Only visible to you:** This is your public tip line. Share the address on your social media profiles, your website, or email signature. Ensuring that someone submitting a message trusts this form belongs to you is critical!

I've also added two new tests:

- One that ensures the submit message form doesn't show up if REQUIRE_PGP is true and the user doesn't have a PGP key, and that it does show the form if the user does have a PGP key.
- One that confirms that the extra fields should up as expected, including links being actually links, and XSS attempts getting escaped.


glenn-sorrentino (changes requested):
1. We need labels above the custom fields in Settings
2. Updated design/content for message form:
Group 28
3. Updated styling for extra fields on message page:
Screenshot 2024-08-22 at 1 47 21 PM

micahflee:
I'm still working on this but here's what I have so far.

Labels on the extra fields:

image

The form is displayed but disabled, in both light mode and dark mode:

image

image

What's left:

- [ ] Fix the display of the extra fields on the profile page. Turns out this is actually quite tricky since we don't know the field names, the number of fields (can be 0, 1, 2, 3, or 4), and the values can get long, but I think I can figure it out.
- [ ] Update the URL from `/profile/{username}` to `/to/{username}`
- [ ] Set up a redirect from `/submit_message/{username}` to `/to/{username}`

migrations/versions/166a3402c391_add_extra_fields_to_user.py
```diff
@@ -0,0 +1,47 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "166a3402c391"
+down_revision = "6e53eac9ea14"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
```
jeremywmoore, line 21:
```suggestion
```

migrations/versions/166a3402c391_add_extra_fields_to_user.py
```diff
@@ -0,0 +1,47 @@
...
+        batch_op.add_column(sa.Column("extra_field_label3", sa.String(), nullable=True))
+        batch_op.add_column(sa.Column("extra_field_value3", sa.String(), nullable=True))
+        batch_op.add_column(sa.Column("extra_field_label4", sa.String(), nullable=True))
+        batch_op.add_column(sa.Column("extra_field_value4", sa.String(), nullable=True))
+
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
```
jeremywmoore, line 36:
```suggestion
```

migrations/versions/166a3402c391_add_extra_fields_to_user.py
```diff
@@ -0,0 +1,47 @@
...
+# revision identifiers, used by Alembic.
+revision = "166a3402c391"
+down_revision = "6e53eac9ea14"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table("users", schema=None) as batch_op:
```
jeremywmoore, line 22: `schema` is already optional, why set it to `None`?
```suggestion
    with op.batch_alter_table("users") as batch_op:
```

hushline/settings.py
```diff
@@ -202,10 +222,23 @@ def index() -> str | Response:
...
 
         # Check if the bio update form was submitted
-        if request.method == "POST" and "update_bio" in request.form:
+        if (
+            request.method == "POST"
+            and "update_bio" in request.form
+            and profile_form.validate_on_submit()
+        ):
             user.bio = request.form["bio"]
+            user.extra_field_label1 = request.form["extra_field_label1"].strip()
```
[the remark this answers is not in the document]
micahflee: Oh yes, good point

migrations/versions/166a3402c391_add_extra_fields_to_user.py
micahflee: Oh, this was part of the autogenerated alembic revision. We actually have some model clean-up to do, so that when we autogenerate a revision it doesn't try to make a bunch of changes. But for now I like your suggestion to remove it.

hushline/routes.py
```diff
@@ -108,8 +108,8 @@ def inbox() -> Response | str:
             is_personal_server=app.config["IS_PERSONAL_SERVER"],
         )
 
-    @app.route("/submit_message/<username>", methods=["GET", "POST"])
-    def submit_message(username: str) -> Response | str:
+    @app.route("/profile/<username>", methods=["GET", "POST"])
+    def profile(username: str) -> Response | str:
```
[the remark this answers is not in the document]
micahflee: Good idea

glenn-sorrentino (approved):

PR #515, opened 2024-08-26
Save progress when the wrong CAPTCHA is failed

### LINKED ISSUE #513: If you fill out the wrong CAPTCHA, you lose your progress
**Describe the bug**
When sending a tip to someone, if you type a message, then fill out the CAPTCHA, and click Send Message, but you filled out the CAPTCHA wrong, it gives you the "Incorrect CAPTCHA. Please try again." error message and redirects you to the user's profile, but the message you typed is lost.

**Expected behavior**
You should get to retry the CAPTCHA without losing your work.

micahflee:
Fixes #513.

This works by storing encrypted data in the session.

When you submit a message and fail the CAPTCHA, it uses `encrypt_field` to encrypt the form's `content_method` and `content` fields and stores them in the session. When loading the profile page again and rebuilding this form, if those encrypted fields are in the session, it decrypts them and adds them to the form, and deletes them from the session.

To demonstrate that it works correctly and doesn't store any plaintext messages in the session cookie, I ran Burp Suite to monitor network requests.

Here's the set-cookie header in my `POST /to/micah`, when I submit a message with an invalid answer for the CAPTCHA:

```
Location: /to/micah
Set-Cookie: session=.eJxdjktvgkAAhP8K2XMxIguoiQeQIPJShPpIY8gCy0NkkWWNReN_L7emncskM5l88wJRdkVdgTsw_3oBjg0GMKUNBR9gTZKGUpwwbqluw6WpjrjtFaMOc4z2HMpRSUbg_D5_gKSjWcSaChMwBymMYSqnKZrJYpYiKYETKEElFjNJkSdQliQ4wVIiD4QasSJCpHtgOgwFOETdPa5LFiUNYShhUY1Z0aRDmau_mkia2hMpKr0047EeEE3nY89s91aZ9vaNbUQndvwVr9VlZXe3TA8Uk881w99bbnbYtT5srO31opfiafx0eiLYNyw_Fou_eEzYPy4-XdzvdWjk5kP3x4K9qYzhYqh4R7QzYGCKYa56m9C32gQ6Qg8xVp8rceVObeytvV65L2-x1rWzT2jbWrw_CbO7yzvVtAi2bUwseAwOVLg3RdY2C_D-AROXhDM.ZszXfg.v-HXWAzcwy4YWlV29aQ9Oo4n8f0; Secure; HttpOnly; Path=/; SameSite=Lax
```

Decoded, the content of that session cookie is:

```json
{
    "_flashes": [
        {
            " t": [
                "error",
                "Incorrect CAPTCHA. Please try again."
            ]
        }
    ],
    "csrf_token": "d4b4d6dda963fda5c424547b3f5762465542e5c6",
    "math_answer": "14",
    "submit_contact_method": "gAAAAAAAAAAA25BAyn5_iNdf-eDSnBD-bNHqVJidyKptO3LbLQG-BmikKspfDS7H-gBFQVJMfWRqQ4oJPljDi3Y0zLyn1Kpe6w==",
    "submit_content": "gAAAAAAAAAAAeYjMxITFgHwDQ01KOkFct_T7NXaRF4SH3TgANOTQJqc4L1y4eeAzG3GM8KeNINy7uCpbBsq9U4KKBbVY19uM-Lk8hSPqbnJ4XSWr1uohfqo="
}
```

As you can see, `submit_contact_method` and `submit_content` are encrypted.

The POST request also redirects to `/to/micah`, which triggers a GET request.

Here's the set-cookie header in my `GET /to/micah` request:

```
Set-Cookie: session=eyJjc3JmX3Rva2VuIjoiZDRiNGQ2ZGRhOTYzZmRhNWM0MjQ1NDdiM2Y1NzYyNDY1NTQyZTVjNiIsIm1hdGhfYW5zd2VyIjoiNiJ9.ZszXfg.ajVMVGvC83zcbwHiZ6_CZuAfg0c; Secure; HttpOnly; Path=/; SameSite=Lax
```

This decodes to:

```json
{
    "csrf_token": "d4b4d6dda963fda5c424547b3f5762465542e5c6",
    "math_answer": "6"
}
```

Immediately on the next page, it deletes the content from the cookie.

And, importantly, the user doesn't lose the progress of the message they were writing, and can just try solving the CAPTCHA again.


hushline/routes.py
```diff
@@ -154,6 +168,10 @@ def submit_message(username: str) -> Response | str:
 
             captcha_answer = request.form.get("captcha_answer", "")
             if not validate_captcha(captcha_answer):
+                # Encrypt the message and store it in the session
+                session["submit_contact_method"] = encrypt_field(form.contact_method.data)
```
jeremywmoore, line 172: I don't love the idea of using the same encryption key for data shared externally as we do for data in the db. Especially since in this instance, users can essentially exercise our encryption library with any chosen plaintext.  At minimum we should probably use a separate key for encrypting this data vs the data in the db.

micahflee:
I've now made it so that the `contact_method` and `content` fields are encrypted in the session with a separate key than the database fields.

I updated the `encrypt_field` and `decrypt_field` to optionally accept a `scope` argument. If you pass in a scope, it derives a new key based on the original key along with the scope, using SHAKE-256 (which I picked simply because @rmlibre was using it for key derivation).

In the case of these two fields in the session, they're encrypted and decrypted with the `submit_message` scope.

hushline/crypto.py
```diff
@@ -1,21 +1,72 @@
...
+        encryption_key_bytes = urlsafe_b64decode(encryption_key)
+        salt_bytes = urlsafe_b64decode(salt)
+
+        # Use Scrypt to derive a unique encryption key based on the scope
+        kdf = Scrypt(
+            salt=salt_bytes,
+            length=32,
+            n=2**14,
+            r=8,
+            p=1,
```
jeremywmoore, line 44: should we extract these literals to constants?

hushline/routes.py
```diff
@@ -154,6 +175,15 @@ def submit_message(username: str) -> Response | str:
 
             captcha_answer = request.form.get("captcha_answer", "")
             if not validate_captcha(captcha_answer):
+                # Encrypt the message and store it in the session
+                scope = "submit_message"
+                salt = generate_salt()
```
jeremywmoore, line 180: I like that part of the secret here lives only in the session, but do we still need a scope if we have a unique salt? My intuition says that the salt binds the scope tighter than the "submit_message" scope.

rmlibre: @jeremywmoore In a situation where the salt is guaranteed to only be used in one place, that may be enough. But, it can get complicated to make that determination. For instance, it's being injected into the session, so access to it is given to other parts of the system (and potentially users). As a result, we can 1. carefully make sure nothing uses it in any other situation; 2. ensure that it can't be harmful if it can somehow be injected into other contexts; or, 3. encode its purpose as a scope. The explicit scope is a much easier, more localized, and less fragile design choice, since it removes the cognitive burden of all that tracking, double-checking, and repeat security analyses when new features are added.

Here are some related resources on the topic:
https://www.cs.cornell.edu/andru/papers/iflow-tosem.pdf
https://www.rfc-editor.org/rfc/rfc1457

hushline/routes.py
```diff
@@ -118,6 +118,27 @@ def profile(username: str) -> Response | str:
...
+            f"{scope}:salt" in session
+            and f"{scope}:contact_method" in session
+            and f"{scope}:content" in session
+        ):
+            try:
+                form.contact_method.data = decrypt_field(
+                    session[f"{scope}:contact_method"], scope, session[f"{scope}:salt"]
+                )
+                form.content.data = decrypt_field(
+                    session[f"{scope}:content"], scope, session[f"{scope}:salt"]
```
rmlibre, line 133: Let's make sure the csrf token is also being used in the key derivation. The salt & csrf_token combination should be an adequately unique & session-bound value.

hushline/routes.py
jeremywmoore: Thanks for the explanation @rmlibre! This feels like sufficient reason to keep the scope in addition to the salt.

hushline/routes.py
micahflee: The csrf_token isn't possible here actually, not without refactoring how this works. It's because we have two separate routes, loading the profile page:

```py
@app.route("/to/<username>", methods=["GET"])
def profile(username: str) -> Response | str:
    form = MessageForm()
```

And submitting a message:

```py
@app.route("/to/<username>", methods=["POST"])
def submit_message(username: str) -> Response | str:
    form = MessageForm()
```

When the user submits a message with an invalid CAPTCHA, it adds stuff to the session and then redirects:

```py
captcha_answer = request.form.get("captcha_answer", "")
if not validate_captcha(captcha_answer):
    # Encrypt the message and store it in the session
    scope = "submit_message"
    salt = generate_salt()
    session[f"{scope}:contact_method"] = encrypt_field(
        form.contact_method.data, scope, salt
    )
    session[f"{scope}:content"] = encrypt_field(form.content.data, scope, salt)
    session[f"{scope}:salt"] = salt

    return redirect(url_for("profile", username=username))
```

When the user subsequently loads the `profile` route, it creates a new instance of `MessageForm()`, which has a new CSRF token that's different from the previous one. So basically, at the point in time where we're ready to decrypt the fields, we've lost access to the original CSRF token.

That said, I think just the fact that each time the CAPTCHA is entered incorrectly it generates a new 128-bit salt that's used just for that scope should be fine for this use-case.

Alternatively, we could give anonymous users some sort of random session ID when the session cookie is first set and include that in the KDF, but it will also need to be stored in the session cookie, and I don't see how that's better than the unique salt.

hushline/crypto.py
micahflee: Implemented in https://github.com/scidsg/hushline/pull/515/commits/740aa15eded55e3d699124a6a2752b77cf4eb92b

micahflee:
@rmlibre Can you take a look at https://github.com/scidsg/hushline/pull/515/commits/601a729400e85df54444f4c2d17cec0a4fabd22e and see if it's a safer way to concatenate items to pass into the KDF?

hushline/routes.py
rmlibre: @micahflee Ok, I see. That makes sense. Ideally there would be some kind of binding information to a particular session. The thing is, using the same key for everything, especially in a multi-user setting, requires that there be very careful and clear domain separation. Another approach may in fact be to actually have distinct keys for each purpose. Though, that may require more thought and potentially structural changes. Key derivation should be fine for now if done carefully.

Apart from that, but what I also think would be sufficient in this case, is to increase the salt size to 32-bytes. The reasoning is that the key in question is a static 32 byte value. When the salt is 16 bytes, there's a max of 2<sup>128</sup> possible different keys that can be output from the key derivation. The number should be closer to 2<sup>256</sup>, otherwise much of the useful entropy from the key is inaccessible. Pseudo-randomly selecting a key from that larger space also decreases the likelihood of a repeat key to comfortable levels (a probability of 2<sup>-85.33...</sup> after about 2<sup>85.33...</sup> encryptions).

rmlibre:
> @rmlibre Can you take a look at [601a729](https://github.com/scidsg/hushline/commit/601a729400e85df54444f4c2d17cec0a4fabd22e) and see if it's a safer way to concatenate items to pass into the KDF?

That looks good 👍🏾

jeremywmoore (approved):

PR #523, opened 2024-08-27
Proton docs

glenn-sorrentino:
- Update ignore file to include npm directory
- Add proton to docs
- Update lock file with prettier

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ hushline/static/data/users_directory.json
 volumes/app
 volumes/redis
 volumes/postgres
+/node_modules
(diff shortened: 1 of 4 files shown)


micahflee (changes requested):
Sorry, just one small suggestion before it's ready. In the Mailvelope instructions you missed the "Send a Message" step.

"Step 3: Add your key to Hush Line" includes text instructions for adding your key, but the screenshots are for sending a message, not adding your key.

Under "Step 3: Add your key to Hush Line", keep the text the same but add new screenshots for adding your key.

Then create a new "Step 4: Send a Message", with the text same text as the Proton step uses:

> Click on Profile in the Hush Line global navigation. Enter a message into your form, and when you submit it, you'll see it encrypted in your browser before it sends - this ensures your message is end-to-end encrypted!

And this one can have the same screenshot that's currently used for Step 3. 

Finally, Step 4 and Step 5 should be renamed to Step 5 and Step 6.

glenn-sorrentino:
Good catches! Should be all updated now.

micahflee (approved):

PR #529, opened 2024-08-29
Add Verified URLs 

glenn-sorrentino:
When a user adds a link back to their Hush Line profile with the `rel="me"` attribute, a verified badge is added to the appropriate URL.

Example shown on my personal portfolio. In this screenshot you see a test link to my admin test account `https://tips.hushline.app/to/admin`. 

Screenshot 2024-08-28 at 5 20 52 PM

On the server-side, we validate address submissions for URLs, and limit them to checking against `tips.hushline.app` addresses.

```python
                # Trigger the rel=me verification for each URL field
                for i in range(1, 5):
                    url_to_verify = getattr(user, f"extra_field_value{i}", "").strip()
                    if url_to_verify:
                        try:
                            response = requests.get(url_to_verify, timeout=5)
                            response.raise_for_status()
                            profile_url = f"https://tips.hushline.app/to/{user.primary_username}"
                            if re.search(
                                rf'<a[^>]+href="{re.escape(profile_url)}"[^>]*rel="me"[^>]*>',
                                response.text,
                            ):
                                setattr(user, f"extra_field_verified{i}", True)
                            else:
                                setattr(user, f"extra_field_verified{i}", False)
                        except requests.exceptions.RequestException as e:
                            current_app.logger.error(f"Error fetching URL for field {i}: {e}")
                            setattr(user, f"extra_field_verified{i}", False)

                db.session.commit()
```

When the conditions validate, a new non-emoji icon appears in the website's fields in Settings and the Profile:

Screenshot 2024-08-28 at 5 20 28 PM

Screenshot 2024-08-28 at 5 28 29 PM

We also add new fields to the DB to track verification status per field:

```python
    extra_field_verified1 = db.Column(db.Boolean, default=False)
    extra_field_verified2 = db.Column(db.Boolean, default=False)
    extra_field_verified3 = db.Column(db.Boolean, default=False)
    extra_field_verified4 = db.Column(db.Boolean, default=False)
```


hushline/settings.py
```diff
@@ -224,57 +224,55 @@ def index() -> str | Response:
...
+                user.extra_field_label1 = profile_form.extra_field_label1.data.strip()
+                user.extra_field_value1 = profile_form.extra_field_value1.data.strip()
+                user.extra_field_label2 = profile_form.extra_field_label2.data.strip()
+                user.extra_field_value2 = profile_form.extra_field_value2.data.strip()
+                user.extra_field_label3 = profile_form.extra_field_label3.data.strip()
+                user.extra_field_value3 = profile_form.extra_field_value3.data.strip()
+                user.extra_field_label4 = profile_form.extra_field_label4.data.strip()
+                user.extra_field_value4 = profile_form.extra_field_value4.data.strip()
+
+                # Trigger the rel=me verification for each URL field
```
jeremywmoore, line 241: I think there's a clear case to be made that url verification would benefit from either parallelization (async coroutines) or deferred processing (celery workers). With 4 urls to verify, using coroutines to verify each url in parallel would roughly reduce the time by a factor of 4.

A deferred execution approach using something like celery would allow us to enqueue a verification task which would update the database with verification status, moving that work entirely outside the time the user is waiting for the next page to load.

There is a significant difference in the complexity of these two approaches. The parallelization approach would require adding an the `async` extra package to flask (https://flask.palletsprojects.com/en/3.0.x/async-await/) and some minor code changes. Adopting celery would add more complexity requiring an additional subsystem like Redis to act as a task broker and result store. 

I'd probably start with the async option here.

hushline/settings.py
```diff
@@ -224,57 +224,55 @@ def index() -> str | Response:
...
+                for i in range(1, 5):
+                    url_to_verify = getattr(user, f"extra_field_value{i}", "").strip()
```
jeremywmoore, line 243: are these all guaranteed to be valid URLs? Should we do an initial filter to only verify fields that are valid urls using https?

jeremywmoore, line 243: We could skip verifying a url if the field value hasn't been changed and only verify a url if the value of the field is changed by the request. If a user wanted a field to be re-verified, they would need to remove it and add it back to their profile.

An alternative to checking whether or not the url has change would be representing the verified column as an enum with states like `Verified`, `Unverified`, `Unknown` where the first two are only set after checking whether the link was found or not, and the third option indicates that no attempt has yet been made to verify (which would be the default value).

hushline/settings.py
```diff
@@ -683,4 +681,26 @@ def delete_account() -> Response | str:
...
+    @authentication_required
+    def verify_rel_me(field_index: int) -> Response:
+        user_id = session.get("user_id")
+        if not user_id:
+            return redirect(url_for("login"))
+
+        user = db.session.get(User, user_id)
+        if not user:
+            flash("User not found.")
+            return redirect(url_for("login"))
```
jeremywmoore, line 694: I think we could incorporate this user lookup into the `authentication_required` decorator. If the user is logged in, but the user doesn't exist, we should clear the session and redirect to login wherever authentication is required.

hushline/settings.py
```diff
@@ -683,4 +681,26 @@ def delete_account() -> Response | str:
         flash("User not found. Please log in again.")
         return redirect(url_for("login"))
 
+    @bp.route("/verify-rel-me/<int:field_index>", methods=["POST"])
```
jeremywmoore, line 684: is this route used? It looked like the main settings route is responsible for verifying links in extra fields.

hushline/settings.py
```diff
@@ -683,4 +681,26 @@ def delete_account() -> Response | str:
...
+        user_id = session.get("user_id")
+        if not user_id:
+            return redirect(url_for("login"))
+
+        user = db.session.get(User, user_id)
+        if not user:
+            flash("User not found.")
+            return redirect(url_for("login"))
+
+        url_to_verify = request.form.get(f"extra_field_value{field_index}", "").strip()
```
jeremywmoore, line 696: `field_index` comes directly from the url, what happens if someone specifies an index of 100?

migrations/versions/83a6b3b09eca_add_verified_fields.py
```diff
@@ -0,0 +1,63 @@
...
+revision = "83a6b3b09eca"
+down_revision = "166a3402c391"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table("authentication_logs", schema=None) as batch_op:
+        batch_op.drop_index("idx_authentication_logs_user_id_timestamp_successful")
```
jeremywmoore, line 23: We don't want to drop this index
```suggestion
```

migrations/versions/83a6b3b09eca_add_verified_fields.py
```diff
@@ -0,0 +1,63 @@
...
+
+    with op.batch_alter_table("message", schema=None) as batch_op:
+        batch_op.alter_column("content", existing_type=sa.TEXT(), nullable=False)
+
+    with op.batch_alter_table("authentication_logs", schema=None) as batch_op:
+        batch_op.create_index(
+            "idx_authentication_logs_user_id_timestamp_successful",
+            ["user_id", "timestamp", "successful"],
+            unique=False,
+        )
```
jeremywmoore, line 61:
```suggestion
```

hushline/settings.py
jeremywmoore: @micahflee would value your take on this

migrations/versions/83a6b3b09eca_add_verified_fields.py
```diff
@@ -0,0 +1,55 @@
...
+    with op.batch_alter_table("message", schema=None) as batch_op:
+        batch_op.alter_column("content", existing_type=sa.TEXT(), nullable=True)
+
+    with op.batch_alter_table("users", schema=None) as batch_op:
+        batch_op.add_column(sa.Column("extra_field_verified1", sa.Boolean(), nullable=True))
+        batch_op.add_column(sa.Column("extra_field_verified2", sa.Boolean(), nullable=True))
+        batch_op.add_column(sa.Column("extra_field_verified3", sa.Boolean(), nullable=True))
+        batch_op.add_column(sa.Column("extra_field_verified4", sa.Boolean(), nullable=True))
+        batch_op.alter_column("password_hash", existing_type=sa.VARCHAR(length=512), nullable=False)
+        batch_op.alter_column("is_verified", existing_type=sa.BOOLEAN(), nullable=False)
```
micahflee, line 32: The auto-genereated migration includes a bunch of stuff unrelated to this PR. That's what #530 is for, to make it so these auto-generated migrations just include what's changed. Can you strip this down to only include the extra_field_verified columns on the user table and just delete the rest of the lines?

hushline/settings.py
micahflee: Yes I agree. At some point it seems inevitable that we'll need some sort of job queue. But for now, I think doing these async is a good approach.

hushline/settings.py
micahflee: Yes I agree. We should check if the field value starts with `https://`, which is the same logic I used in the template to determine if the fields should be links or not:

```jinja2
{% if field.value.startswith('https://') %}
```

hushline/settings.py
```diff
@@ -224,57 +224,55 @@ def index() -> str | Response:
...
+                user.extra_field_value3 = profile_form.extra_field_value3.data.strip()
+                user.extra_field_label4 = profile_form.extra_field_label4.data.strip()
+                user.extra_field_value4 = profile_form.extra_field_value4.data.strip()
+
+                # Trigger the rel=me verification for each URL field
+                for i in range(1, 5):
+                    url_to_verify = getattr(user, f"extra_field_value{i}", "").strip()
+                    if url_to_verify:
+                        try:
+                            response = requests.get(url_to_verify, timeout=5)
```
micahflee, line 246: Yeah with a 5 second timeout and 4 URLs to verify, if you add 4 totally invalid URLs that don't respond, then it will take 20 seconds to save the page. Async would definitely be good, meaning it would take 5 seconds.

hushline/settings.py
```diff
@@ -224,57 +224,55 @@ def index() -> str | Response:
...
+                            response.raise_for_status()
+                            profile_url = f"https://tips.hushline.app/to/{user.primary_username}"
+                            if re.search(
```
micahflee, line 249: Instead of searching for the rel="me" link using regex, it would be better to use an HTML parsing library like BeautifulSoup (the python module is bs4, and we actually already use bs4 as a dev dependency because I used it in tests, but we could move it to a main dependency).

The reason is because there's a lot of different ways that the link could be valid HTML but the regex wouldn't notice it. For example, these are all valid:

```html
<a href="https://tips.hushline.app/to/admin" rel="nofollow me">
<a href='https://tips.hushline.app/to/admin' rel='me'>
<a href="https://tips.hushline.app/to/admin" rel=me>
<a
    href = "https://tips.hushline.app/to/admin"
    rel = "me"
>
<  a       href   =      "https://tips.hushline.app/to/admin" rel                      =                                         "me"    >
```

(If you add a link to your github profile, it adds `rel="nofollow me"`.)

Here's how you can use bs4:

```py
import requests
import bs4

def verify(field_value_url, profile_url):
    r = requests.get(field_value_url)
    soup = bs4.BeautifulSoup(r.text, "html.parser")
    for link in soup.find_all("a"):
        if link.get("href") == profile_url and "me" in link.get("rel"):
            print("Found rel=me link to profile URL")

verify("https://glennsorrentino.com/", "https://tips.hushline.app/to/admin")
verify("https://github.com/micahflee", "https://infosec.exchange/@micahflee")
```

hushline/settings.py
micahflee: I don't think this route is being used. But also it doesn't do any verification, so it should probably be removed.

tests/test_profile.py
```diff
@@ -155,12 +155,13 @@ def test_profile_extra_fields(client: FlaskClient, app: Flask) -> None:
     assert b"Signal username" in response.data
     assert b"singleusername.666" in response.data
     assert b"Arbitrary Link" in response.data
-    # URLs should turn into links
-    assert (
-        '<a href="https://scidsg.org/" target="_blank" rel="noopener noreferrer"'
-        "\n                  >https://scidsg.org/</a\n                >" in response.text
-    )
+
+    # URLs should turn into links - check that critical parts of the link are present
```
micahflee, line 159: The problem with how this is checking for specific parts of the link being present is that if any link on the page includes `target="_blank"` then this test will pass, even if this specific link doesn't have that. Instead you can use bs4 to find this specific link, and then make sure it has the properties you want it to have.

micahflee (changes requested):

hushline/settings.py
```diff
@@ -224,57 +224,55 @@ def index() -> str | Response:
...
-            user.extra_field_label4 = profile_form.extra_field_label4.data.strip()
-            user.extra_field_value4 = profile_form.extra_field_value4.data.strip()
-            db.session.commit()
-            flash("👍 Bio updated successfully.")
-            return redirect(url_for("settings.index"))
+        # Handle form submissions
+        if request.method == "POST":
+            # Update bio and custom fields
+            if "update_bio" in request.form and profile_form.validate_on_submit():
+                user.bio = request.form["bio"].strip()
```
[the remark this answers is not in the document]
glenn-sorrentino: Changes in #97ba5a2

hushline/settings.py
```diff
@@ -224,57 +224,55 @@ def index() -> str | Response:
...
+                user.extra_field_label1 = profile_form.extra_field_label1.data.strip()
+                user.extra_field_value1 = profile_form.extra_field_value1.data.strip()
+                user.extra_field_label2 = profile_form.extra_field_label2.data.strip()
+                user.extra_field_value2 = profile_form.extra_field_value2.data.strip()
+                user.extra_field_label3 = profile_form.extra_field_label3.data.strip()
+                user.extra_field_value3 = profile_form.extra_field_value3.data.strip()
+                user.extra_field_label4 = profile_form.extra_field_label4.data.strip()
+                user.extra_field_value4 = profile_form.extra_field_value4.data.strip()
```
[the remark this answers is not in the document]
glenn-sorrentino: Changes in a28e0fc

hushline/settings.py
glenn-sorrentino: Changes in 90fadf7

hushline/settings.py
glenn-sorrentino: Changes in c114b3a

hushline/settings.py
glenn-sorrentino: Changes in 83a47e9

Confirmed in logs:
```
2024-08-30 11:12:05,440:ERROR:Error fetching URL for field 1: glenn.01
2024-08-30 11:12:05,440:ERROR:Error fetching URL for field 2: sciencedesign
2024-08-30 11:12:05,440:ERROR:Error fetching URL for field 3: fosstodon.org/@scidsg
2024-08-30 11:12:05,949:INFO:127.0.0.1 - - [30/Aug/2024 11:12:05] "POST /settings/ HTTP/1.1" 302 -
2024-08-30 11:12:05,959:INFO:127.0.0.1 - - [30/Aug/2024 11:12:05] "GET /settings/ HTTP/1.1" 200 -
2024-08-30 11:12:05,995:INFO:127.0.0.1 - - [30/Aug/2024 11:12:05] "GET /static/css/style.css HTTP/1.1" 304 -
2024-08-30 11:12:05,997:INFO:127.0.0.1 - - [30/Aug/2024 11:12:05] "GET /static/js/global.js HTTP/1.1" 304 -
2024-08-30 11:12:05,998:INFO:127.0.0.1 - - [30/Aug/2024 11:12:05] "GET /static/js/settings.js HTTP/1.1" 304 -
2024-08-30 11:12:06,000:INFO:127.0.0.1 - - [30/Aug/2024 11:12:06] "GET /static/img/app/dropdown.png HTTP/1.1" 304 -
2024-08-30 11:12:06,001:INFO:127.0.0.1 - - [30/Aug/2024 11:12:06] "GET /static/img/app/icon-search.png HTTP/1.1" 304 -
2024-08-30 11:12:06,007:INFO:127.0.0.1 - - [30/Aug/2024 11:12:06] "GET /static/img/app/icon-verified-lm.png HTTP/1.1" 304 -
2024-08-30 11:12:06,032:INFO:127.0.0.1 - - [30/Aug/2024 11:12:06] "GET /static/css/style.css HTTP/1.1" 304 -
2024-08-30 11:15:44,992:INFO: * Detected change in '/Users/glennsorrentino/Library/Mobile Documents/com~apple~CloudDocs/Git/hushline/hushline/settings.py', reloading
2024-08-30 11:15:45,109:INFO: * Restarting with stat
2024-08-30 11:15:45,686:WARNING: * Debugger is active!
2024-08-30 11:15:45,694:INFO: * Debugger PIN: 328-910-462
2024-08-30 11:16:20,111:INFO:127.0.0.1 - - [30/Aug/2024 11:16:20] "GET /settings/ HTTP/1.1" 200 -
2024-08-30 11:16:20,138:INFO:127.0.0.1 - - [30/Aug/2024 11:16:20] "GET /static/js/global.js HTTP/1.1" 304 -
2024-08-30 11:16:20,139:INFO:127.0.0.1 - - [30/Aug/2024 11:16:20] "GET /static/css/style.css HTTP/1.1" 304 -
2024-08-30 11:16:20,139:INFO:127.0.0.1 - - [30/Aug/2024 11:16:20] "GET /static/js/settings.js HTTP/1.1" 304 -
2024-08-30 11:16:20,143:INFO:127.0.0.1 - - [30/Aug/2024 11:16:20] "GET /static/img/app/dropdown.png HTTP/1.1" 304 -
2024-08-30 11:16:20,143:INFO:127.0.0.1 - - [30/Aug/2024 11:16:20] "GET /static/img/app/icon-search.png HTTP/1.1" 304 -
2024-08-30 11:16:20,156:INFO:127.0.0.1 - - [30/Aug/2024 11:16:20] "GET /static/img/app/icon-verified-lm.png HTTP/1.1" 304 -
2024-08-30 11:16:24,087:INFO:127.0.0.1 - - [30/Aug/2024 11:16:24] "POST /settings/ HTTP/1.1" 302 -
2024-08-30 11:16:24,102:INFO:127.0.0.1 - - [30/Aug/2024 11:16:24] "GET /settings/ HTTP/1.1" 200 -
2024-08-30 11:16:24,128:INFO:127.0.0.1 - - [30/Aug/2024 11:16:24] "GET /static/css/style.css HTTP/1.1" 304 -
2024-08-30 11:16:24,143:INFO:127.0.0.1 - - [30/Aug/2024 11:16:24] "GET /static/img/app/dropdown.png HTTP/1.1" 304 -
2024-08-30 11:16:24,145:INFO:127.0.0.1 - - [30/Aug/2024 11:16:24] "GET /static/js/global.js HTTP/1.1" 304 -
2024-08-30 11:16:24,146:INFO:127.0.0.1 - - [30/Aug/2024 11:16:24] "GET /static/js/settings.js HTTP/1.1" 304 -
2024-08-30 11:16:24,148:INFO:127.0.0.1 - - [30/Aug/2024 11:16:24] "GET /static/img/app/icon-verified-lm.png HTTP/1.1" 304 -
2024-08-30 11:16:24,151:INFO:127.0.0.1 - - [30/Aug/2024 11:16:24] "GET /static/img/app/icon-search.png HTTP/1.1" 304 -
```

hushline/settings.py
glenn-sorrentino: Removed unused route in c114b3a

hushline/settings.py
glenn-sorrentino: Removed this unused route in c114b3a

hushline/settings.py
glenn-sorrentino: Fixed in 90fadf7

hushline/settings.py
glenn-sorrentino: I think we should probably check URLs that have already been verified. A user could change their website at any time and we'd want to make sure the checks reflect that.

hushline/settings.py
glenn-sorrentino: Changes in bcae3d3

migrations/versions/83a6b3b09eca_add_verified_fields.py
glenn-sorrentino: Changes in 32b7af2

hushline/model.py
```diff
@@ -62,6 +62,10 @@ class User(Model):
     extra_field_value3: Mapped[Optional[str]]
     extra_field_label4: Mapped[Optional[str]]
     extra_field_value4: Mapped[Optional[str]]
+    extra_field_verified1: Mapped[Optional[bool]] = mapped_column(default=False)
+    extra_field_verified2 = db.Column(db.Boolean, default=False)
```
jeremywmoore, line 66: Please update the rest of these like the first one

hushline/settings.py
```diff
@@ -224,57 +273,62 @@ def index() -> str | Response:
...
-            db.session.commit()
-            flash("👍 Directory visibility updated successfully.")
-            return redirect(url_for("settings.index"))
+        # Handle form submissions
+        if request.method == "POST":
+            # Update bio and custom fields
+            if "update_bio" in request.form and profile_form.validate_on_submit():
+                user.bio = profile_form.bio.data
+
+                async def perform_verification() -> None:
```
jeremywmoore, line 282: To get the benefits of parallelism we'll need to schedule an async task for each url.

jeremywmoore (changes requested):

hushline/settings.py
```diff
@@ -224,57 +273,65 @@ def index() -> str | Response:
...
+                                # Schedule async task for verification
+                                task = verify_url(session, user, i, url_to_verify, profile_url)
+                                tasks.append(task)
+
+                        # Run all the tasks concurrently
+                        if tasks:  # Only gather if there are tasks to run
+                            await asyncio.gather(*tasks)
+
+                # Run the async verification function
+                asyncio.run(perform_verification())
```
jeremywmoore, line 319: I don't think this needs to be an async function if we change the signature of the route function to be async because flask should invoke the route handler, see the snippet at the top of https://flask.palletsprojects.com/en/3.0.x/async-await/

hushline/settings.py
```diff
@@ -224,57 +273,65 @@ def index() -> str | Response:
...
+                            # If the value is empty, reset the verification status
+                            if not value.strip():
+                                setattr(user, f"extra_field_verified{i}", False)
+                                continue
+
+                            # Verify the URL only if it starts with "https://"
+                            url_to_verify = value.strip()
+                            if url_to_verify.startswith("https://"):
+                                profile_url = (
+                                    f"https://tips.hushline.app/to/{user.primary_username}"
```
jeremywmoore, line 308: Are verified URLs a feature limited to tips.hushline.app?

hushline/settings.py
```diff
@@ -224,57 +273,55 @@ def index() -> str | Response:
...
-            db.session.commit()
-            flash("👍 Directory visibility updated successfully.")
-            return redirect(url_for("settings.index"))
+        # Handle form submissions
+        if request.method == "POST":
+            # Update bio and custom fields
+            if "update_bio" in request.form and profile_form.validate_on_submit():
+                user.bio = profile_form.bio.data
+
+                base_url = current_app.config["HUSHLINE_TIPS_URL"]
```
jeremywmoore, line 282: I think we can use Flask to generate this url for us:
```suggestion
                profile_url = url_for("profile", _external=True, username=user.primary_username)
```
Note that we need to configure the `SERVER_NAME` config variable with the value of the environment variable.

hushline/settings.py
```diff
@@ -224,57 +273,55 @@ def index() -> str | Response:
...
+                        value_field = getattr(profile_form, f"extra_field_value{i}", "")
+
+                        label = label_field.data if hasattr(label_field, "data") else label_field
+                        setattr(user, f"extra_field_label{i}", label)
+
+                        value = value_field.data if hasattr(value_field, "data") else value_field
+                        setattr(user, f"extra_field_value{i}", value)
+
+                        # If the value is empty, reset the verification status
+                        if not value.strip():
```
jeremywmoore, line 297: Values are already stripped by the form filter
```suggestion
                        if not value:
```

hushline/settings.py
```diff
@@ -224,57 +273,55 @@ def index() -> str | Response:
...
+                            setattr(user, f"extra_field_verified{i}", False)
+                            continue
+
+                        # Verify the URL only if it starts with "https://"
+                        url_to_verify = value.strip()
```
jeremywmoore, line 302: should be already stripped
```suggestion
                        url_to_verify = value
```

jeremywmoore (changes requested):

hushline/settings.py
```diff
@@ -224,57 +273,58 @@ def index() -> str | Response:
...
+                    tasks = []
+                    for i in range(1, 5):
+                        label_field = getattr(profile_form, f"extra_field_label{i}", "")
+                        value_field = getattr(profile_form, f"extra_field_value{i}", "")
+
+                        label = label_field.data if hasattr(label_field, "data") else label_field
+                        setattr(user, f"extra_field_label{i}", label)
+
+                        value = value_field.data if hasattr(value_field, "data") else value_field
+                        setattr(user, f"extra_field_value{i}", value)
```
jeremywmoore, line 297: Should we skip the url verification if the field value hasn't changed?

jeremywmoore (dismissed):
Looks good, we will need a change in our infra repo to set the `SERVER_NAME` variable before this will work.

micahflee (changes requested):
I tested it and it looks great, but I think this should include one small change.

Our extra fields links themselves should include `rel="me"`! Right now it's `rel="noopener noreferrer"` but we should add a `me` to that so that people can e.g. link to their Hush Line profile from Mastodon or other services that use it, and those will be marked as verified links on those services.

And of course, the whole SERVER_NAME thing. I tried to deploy a test server and it broke because of this. So the infra side needs to be fixed first. And the dev deploys in particular could tricky... @jeremywmoore ?

glenn-sorrentino:
> I tested it and it looks great, but I think this should include one small change.
> 
> Our extra fields links themselves should include `rel="me"`! Right now it's `rel="noopener noreferrer"` but we should add a `me` to that so that people can e.g. link to their Hush Line profile from Mastodon or other services that use it, and those will be marked as verified links on those services.
> 
> And of course, the whole SERVER_NAME thing. I tried to deploy a test server and it broke because of this. So the infra side needs to be fixed first. And the dev deploys in particular could tricky... @jeremywmoore ?

Added in 9d030d3

jeremywmoore:
> > I tested it and it looks great, but I think this should include one small change.
> > 
> > Our extra fields links themselves should include `rel="me"`! Right now it's `rel="noopener noreferrer"` but we should add a `me` to that so that people can e.g. link to their Hush Line profile from Mastodon or other services that use it, and those will be marked as verified links on those services.
> > 
> > And of course, the whole SERVER_NAME thing. I tried to deploy a test server and it broke because of this. So the infra side needs to be fixed first. And the dev deploys in particular could tricky... @jeremywmoore ?
> 
> Added in 9d030d3

I'm working on this, DO can inject the domain name into the app via magic envelope var.

jeremywmoore (approved):

micahflee (approved):

PR #532, opened 2024-08-31
New rules to make small changes easier

glenn-sorrentino:
Adds two new rules for admins:
- Bypass CSS review requirement
- Bypass version file review reqs


jeremywmoore:
I'd recommend not merging automatically but instead approving.

glenn-sorrentino:
> I'd recommend not merging automatically but instead approving.

Yes, good recommendation. It's in d85cb2a.

jeremywmoore:
Does this auto-approve if non-css or non-version changes are included in a PR in addition to CSS or `version.py` changes?

.github/workflows/bypass-css-review.yaml
```diff
@@ -0,0 +1,34 @@
...
+          AUTHOR=$(gh pr view ${{ github.event.pull_request.number }} --json author --jq '.author.login')
+          ADMIN_CHECK=$(gh api user/memberships/orgs/scidsg --jq '.role')
+          if [ "$ADMIN_CHECK" != "admin" ]; then
+            echo "User is not an admin or is not a public member of the organization. Exiting."
+            exit 1
+          fi
+
+      - name: Check for CSS files and approve if conditions are met
+        run: |
+          if git diff --name-only main ${{ github.head_ref }} | grep -q '\.css$'; then
```
jeremywmoore, line 27: I think instead of success on matching files, we should instead fail on non-matching,
```suggestion
          if ! git diff --name-only main ${{ github.head_ref }} | grep -v -q '\.css$'; then
```
This has the effect of auto-approving only when all the files modified are of the desired type.

.github/workflows/bypass-version-review.yaml
```diff
@@ -0,0 +1,34 @@
...
+          AUTHOR=$(gh pr view ${{ github.event.pull_request.number }} --json author --jq '.author.login')
+          ADMIN_CHECK=$(gh api user/memberships/orgs/scidsg --jq '.role')
+          if [ "$ADMIN_CHECK" != "admin" ]; then
+            echo "User is not an admin or is not a public member of the organization. Exiting."
+            exit 1
+          fi
+
+      - name: Check for hushline/version.py and approve if conditions are met
+        run: |
+          if git diff --name-only main ${{ github.head_ref }} | grep -q '^hushline/version.py$'; then
```
jeremywmoore, line 27: Following the same suggestion as above,
```suggestion
          if ! git diff --name-only main ${{ github.head_ref }} | grep -v -q '^hushline/version.py$'; then
```

.github/workflows/bypass-css-review.yaml
glenn-sorrentino: Wouldn't this mean that one workflow will always fail for most PRs?

Screenshot 2024-09-03 at 9 56 10 AM

.github/workflows/bypass-version-review.yaml
glenn-sorrentino: Same as above...

Screenshot 2024-09-03 at 9 56 10 AM

.github/workflows/bypass-css-review.yaml
glenn-sorrentino: Screenshot 2024-09-03 at 9 58 12 AM

glenn-sorrentino: I updated to use a non-failing exit

.github/workflows/bypass-version-review.yaml
glenn-sorrentino: Updated with non-failing exit

.github/workflows/bypass-version-review.yaml
```diff
@@ -0,0 +1,34 @@
+name: Bypass Version File Review by Admin
```
jeremywmoore, line 1: Do we need this workflow anymore? It looks like this case is now handled by `bypass-review.yaml`

glenn-sorrentino: removed

jeremywmoore (approved):

PR #543, opened 2024-09-07
Update docs for managed service

glenn-sorrentino:
There were a lot of old artifacts from yesteryear in the docs, so I cleaned them up by removing unneeded content, updating images, and revising the remaining documentation.

### DIFF
--- a/docs/3-managed-service.md
+++ b/docs/3-managed-service.md
@@ -103,7 +103,7 @@ Users may opt in to being listed on our public directory, making it easy for any
 
 #### Verified Addresses
 
-Verified addresses establishes a relationship between your Hush Line account and another website. Verification is automatic, and all you need to do is include a link to your Hush Line profile containing `rel="me"` on the page of the address you want verified. 
+Verified addresses establishes a relationship between your Hush Line account and another website. Verification is automatic, and all you need to do is include a link to your Hush Line profile containing `rel="me"` on the page of the address you want verified.
 
 For example, if I want to verify my personal website `glennsorrentino.com` I would add this to my page:
 
@@ -111,7 +111,7 @@ For example, if I want to verify my personal website `glennsorrentino.com` I wou
 <a href="https://tips.hushline.app/to/glenn" rel="me">Hush Line</a>
 ```
 
-In your Hush Line settings, when I add `https://glennsorrentino.com` you'll see a verified checkmark. 
+In your Hush Line settings, when I add `https://glennsorrentino.com` you'll see a verified checkmark.
 
 ### Authentication
 
--- a/docs/3-managed-service.md
+++ b/docs/3-managed-service.md
@@ -36,7 +36,7 @@ Our verified accounts feature is designed to ensure that messages reach their in
 
 To ensure your account is recognized as authentic, users belonging to the categories listed above can apply for verification. Follow these steps to initiate the verification process:
 
-1. **First, add a verified URL to your Hush Line profile:** In Hush Line settings you can add up to four additional fields - a URL, Signal username, phone number, or anything else. To verify a URL, you simply need to add a link to your Hush Line address on the page for that URL with `rel="me"` in the link's markup. For example:
+1. **First, add a verified URL to your Hush Line profile:** In Hush Line settings you can add up to four additional fields - a URL, Signal username, phone number, or anything else. To verify a URL, add it to your Hush Line extra fields using `https://` - `https://scidsg.org`, for example. Then, add a link to your Hush Line address on the page for that URL with `rel="me"` in the markup. For example:
 
    ```html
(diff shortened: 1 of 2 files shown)


docs/3-managed-service.md
```diff
@@ -74,88 +65,62 @@ Once your Display Name matches the format above, we'll promptly verify your acco
...
-Users can set a human-readable display name so that someone submitting a message can see "Submit a message to Science & Design" rather than "Submit a message to scidsg".
+For example, if I want to verify my personal website `glennsorrentino.com` I would add this to my page:
 
-##### Verification
+```html
+<a href="https://tips.hushline.app/to/glenn" rel="me">Hush Line</a>
+```
 
-As a security measure, if a user changes their display name after verification, they'll lose that status and must re-verify their account.
+In your Hush Line settings, when I add `https://glennsorrentino.com` you'll see a verified checkmark.
```
jeremywmoore, line 114: I wonder if this is worth a screenshot as users might try adding urls to their bio instead of "Extra Fields".

docs/3-managed-service.md
```diff
@@ -168,50 +133,34 @@ Changing your username is easy but can lead to confusion for end-users, especial
...
-<img src="img/settings.advanced.png">
+![Settings - Email](https://github.com/user-attachments/assets/4a1ec530-cf19-432e-81ee-cbbf53da1261)
 
-## Admin View
-
-Admins can control certain aspects of user settings without having to log in to the server or manually modifying the database. This is where users are verified, where paid features can be manually enabled, or where admins can be created.
+#### Email Delivery
 
-A highlight panel displays the total number of users and how many have enabled 2fa or added a PGP key. This view helps with internal decision-making, like education campaigns for security best practices.
+We use Riseup.net for our email forwarding provider. All you need to do is add an email address and you'll get new secure messages delivered directly to your inbox.
```
jeremywmoore, line 144: It might be worth being explicit here as Hush Line also has its own "inbox"
```suggestion
We use Riseup.net for our email forwarding provider. All you need to do is add an email address and you'll get new secure messages delivered directly to your email inbox.
```

docs/3-managed-service.md
jeremywmoore: Maybe add a note that says "Only urls using https protocol are eligible for verification"?

glenn-sorrentino: Good call. I also updated the criteria for verification. It occurs to me that we can simply use verified URLs as a proxy for authenticity of the receiver. Both inbound to Hush Line, or outbound to Mastodon.

docs/3-managed-service.md
```diff
@@ -4,158 +4,106 @@ The hosted version of Hush Line is the first free and open-source anonymous tip-
...
-<img src="img/verified-account.png">
+![Profile](https://github.com/user-attachments/assets/b34bb0c8-ad9e-4aef-a412-5085c61bb230)
 
 ### Requesting Verification
 
 To ensure your account is recognized as authentic, users belonging to the categories listed above can apply for verification. Follow these steps to initiate the verification process:
 
-1. **Send a Message:** Use the contact form in the app to reach out to us. Include your name or organization's name and a preferred method of contact (email or phone number). We'll get back to you to schedule a verification meeting.
-2. **Prepare Your Documents:** To verify your identity or your authority to represent an organization, please include the relevant documents for your role:
+1. **First, add a verified URL to your Hush Line profile:** In Hush Line settings you can add up to four additional fields - a URL, Signal username, phone number, or anything else. To verify a URL, add it to your Hush Line extra fields using `https://` - `https://scidsg.org`, for example. Then, add a link to your Hush Line address on the page for that URL with `rel="me"` in the markup. For example:
```
jeremywmoore, line 39: Does this mean that we're eliminating the manual verification process? If we're just relying on verified URLs, what's to stop someone from setting up a ring of fake profiles that are all linked to "verify" each other?

docs/3-managed-service.md
```diff
@@ -4,158 +4,106 @@ The hosted version of Hush Line is the first free and open-source anonymous tip-
...
+Users can set a human-readable display name so that someone submitting a message can see "Submit a message to Science & Design" rather than "Submit a message to scidsg". As a security measure, if a user changes their display name after verification, they'll lose that status and must re-verify their account.
 
-Users can tailor Hush Line to their needs, from adding a display name to adding multiple usernames.
+#### Public Directory
 
-<img src="img/settings.profile.png">
+Users may opt in to being listed on our public directory, making it easy for anyone to find their Hush Line account.
 
-#### Display Name
+#### Verified Addresses
```
jeremywmoore, line 89: "Addresses" in this context is confusing

brassy-endomorph:
We did a lot to clean up the docs. Is this still relevant or were the changes already worked in?

PR #545, opened 2024-09-07
Update bypass-review.yaml

glenn-sorrentino:
Check the user's role using memberships API (works for both public and private members)

### DIFF
--- a/.github/workflows/bypass-review.yaml
+++ b/.github/workflows/bypass-review.yaml
@@ -31,11 +31,9 @@ jobs:
           # Check the user's role using memberships API (works for both public and private members)
           ADMIN_CHECK=$(gh api orgs/scidsg/memberships/$AUTHOR --jq '.role')
 
-          # Output the result for debugging purposes
-          echo "Admin check response: $ADMIN_CHECK"
-
+          # If the user isn't an admin, exit with a detailed error message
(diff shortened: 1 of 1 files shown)


.github/workflows/bypass-review.yaml
```diff
@@ -25,8 +25,15 @@ jobs:
...
         run: |
+          # Get the author of the PR
           AUTHOR=$(gh pr view ${{ github.event.pull_request.number }} --json author --jq '.author.login')
-          ADMIN_CHECK=$(gh api orgs/scidsg/members/$AUTHOR --jq '.role')
+
+          # Check the user's role using memberships API (works for both public and private members)
+          ADMIN_CHECK=$(gh api orgs/scidsg/memberships/$AUTHOR --jq '.role')
+
+          # Output the result for debugging purposes
+          echo "Admin check response: $ADMIN_CHECK"
```
jeremywmoore, line 35: Do we need this? What if we included the user's role in the error message if it's not admin?

jeremywmoore (approved):

.github/workflows/bypass-review.yaml
glenn-sorrentino: 👍 Good call. The role is included in the message now. 2d1e152

jeremywmoore (approved):

PR #560, opened 2024-09-09
Update version.py

glenn-sorrentino:


scidsg-dev (approved):

Issue #546, opened 2024-09-09
Postgres major version should be pinned


brassy-endomorph:
We do not mention a recommended or minimum supported Postgres version. This is something admins (internal, external) would like to be aware of.

**Describe the solution you'd like**

Docs should say "use Postgres 16.x" and that version should be reflected in the `docker-compose.yml` in the repo.

**Describe alternatives you've considered**

Not pinning? Letting people just cowboy it any use any ol' version? Not ideal

**Additional context**

This will at the very least give us consistency within the dev team and help make our env a little more reproducible.

PR #572, opened 2024-09-10
Use postgres everywhere

### LINKED ISSUE #557: Test cases should use Postgres and not `sqlite:///:memory:`

There is drift between the test and prod environment, and with something as complex as Postgres and remote state, we don't want that drift. Especially if we ever use any PG specific features

In `conftest.py` we should have a fixture that gives us a fresh PG database with which to test on with `scope="function"`. We can use PG template DBs as a fixture with `scope="session"` to only have to create the tables once and that use that in the fixture the provides the DB to the test cases.

**Additional context**

This may be a blocker for #528 and if it's not, we'd have to likely add a bunch of Python that would only ever be needed for the test cases and not the real app code, and that's some serious code smell IMO.

This is very likely a blocker for meaningfully implementing #555

### LINKED ISSUE #569: Foreign keys are not respected on row deletion

STR (in the dev env):

1. Create a user
2. Open a private window, submit a message
3. Delete that user via `sqlite3` CLI with `delete from users;`
4. Create a new user
5. Login and look at inbox

Expected result: inbox is empty.

Actual result: inbox shows previous message

Why? Step 3 allows deleting rows from the `users` table when the foreign keys should prevent that (or should cascade to delete related rows).

Within SQLite, we can enable the `foreign_key_check` pragma.

Ideally we should just switch to Postgres across the board as per #557 

**Additional context**

This only affects the dev and test environments.

brassy-endomorph:
- Adds a script to start a standalone docker container for testing
- Configures tests to use postgres including fixtures for isolating database per test function
- Updates github actions to test against postgres

Fixes #557
Fixes #569
Partially addresses #546


glenn-sorrentino (approved):
✅ Ran app locally
✅ Created new account
✅ Added settings
✅ Tested with new dev instructions

Issue #533, opened 2024-09-01
Customer Branding


glenn-sorrentino:
Customers want basic customization of instances to align with their brand. 

**Acceptance Criteria**

- Applicable only for Single-tenant instances
- An admin can:
   - Change primary color
      - Automatically applies to buttons, tabs, links, etc. 
   - Change background color (infer a tint from 15% of the primary)
   - Change header H1 

**Design**

![Image](https://github.com/user-attachments/assets/7b227afc-0af0-4a17-8f4f-349a13ad8f3b)

PR #573, opened 2024-09-10
feat: enable customer color theme & name branding #533

rmlibre:
Initial updates needed to implement the requested branding features on Roadmap item #533.

Research Resources:
- https://stripe.com/blog/accessible-color-systems
- https://evilmartians.com/chronicles/oklch-in-css-why-quit-rgb-hsl
- https://atmos.style/blog/lch-color-space
- https://lea.verou.me/blog/2020/04/lch-colors-in-css-what-why-and-how/
- https://dev.to/greenteaisgreat/the-new-css-color-format-you-didnt-know-you-needed-oklch-10hf
- https://oklch.com/#48.55,0.222,304.47,100
- https://webaim.org/resources/contrastchecker/
- https://www.w3.org/TR/WCAG22/#contrast-enhanced

---

# Light Mode

hushline_533_-_light_theme_purple_mobile_-_0006

### Purple Brand Contrast [12.47:1](https://webaim.org/resources/contrastchecker/?fcolor=410083&bcolor=FBF3FF)
hushline_533_-_light_theme_purple_-_0005

### Blue Brand Contrast [14.26:1](https://webaim.org/resources/contrastchecker/?fcolor=000048&bcolor=D9DEEB)
hushline_533_-_light_theme_blue_-_0005

### Pistachio Brand Contrast [10.38:1](https://webaim.org/resources/contrastchecker/?fcolor=1A4603&bcolor=F1FCF2)
hushline_533_-_light_theme_pistachio_-_0005

### Black Brand Contrast [11.09:1](https://webaim.org/resources/contrastchecker/?fcolor=030303&bcolor=BEBEBE)
hushline_533_-_light_theme_black_-_0005

### Pink Brand Contrast [7.69:1](https://webaim.org/resources/contrastchecker/?fcolor=9E0045&bcolor=FFF4F7)
hushline_533_-_light_theme_pink_-_0005

---

# Dark Mode

hushline_533_-_dark_theme_purple_mobile_-_0006

### Purple Brand Contrast [13.05:1](https://webaim.org/resources/contrastchecker/?fcolor=FFCAFF&bcolor=181519)
hushline_533_-_dark_theme_purple_-_0006

### Blue Brand Contrast [14.89:1](https://webaim.org/resources/contrastchecker/?fcolor=C5EAFF&bcolor=101116)
hushline_533_-_dark_theme_blue_-_0006

### Pistachio Brand Contrast [16.91:1](https://webaim.org/resources/contrastchecker/?fcolor=E4FFDA&bcolor=131713)
hushline_533_-_dark_theme_pistachio_-_0006

### Black Brand Contrast [15.28:1](https://webaim.org/resources/contrastchecker/?fcolor=E8E8E8&bcolor=121212)
hushline_533_-_dark_theme_black_-_0006

### Pink Brand Contrast [11.88:1](https://webaim.org/resources/contrastchecker/?fcolor=FFBAFF&bcolor=1A1415)
hushline_533_-_dark_theme_pink_-_0006


hushline/settings.py
```diff
@@ -194,6 +194,14 @@ class ProfileForm(FlaskForm):
     )
 
 
+class UpdateBrandPrimaryColorForm(FlaskForm):
+    hex_color = StringField("HEX Color", validators=[DataRequired(), Length(min=3, max=7)])
```
jeremywmoore, line 198: Should we use a regex validator on this field? https://stackoverflow.com/a/1636354

jeremywmoore (commented):
Hey @rmlibre, thanks for pushing up this draft! I took a pass through and made a few suggestions.

hushline/settings.py
rmlibre: Addressed by (https://github.com/scidsg/hushline/pull/573/commits/75269c70e3cbd667e271e8d7e3be12a0118ca6ed, https://github.com/scidsg/hushline/pull/573/commits/5f246a8ea6caeb6b8559f562d878e68e8ce9a417)

hushline/settings.py
```diff
@@ -704,6 +712,50 @@ def update_smtp_settings() -> Response | str:
         flash("👍 SMTP settings updated successfully")
         return redirect(url_for(".index"))
 
+    @bp.route("/update-brand-primary-color", methods=["POST"])
+    @authentication_required
```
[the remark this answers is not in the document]
rmlibre: Addressed by (https://github.com/scidsg/hushline/pull/573/commits/d847a509ef134cc86cea4dc00a4778d97e73d579)

hushline/settings.py
```diff
@@ -704,6 +712,50 @@ def update_smtp_settings() -> Response | str:
...
+                user.hex_color = form.hex_color
+                db.session.commit()
+                flash("👍 Brand primary color updated successfully.")
+                return redirect(url_for(".index"))
+
+        flash("User not found. Please log in again.")
+        return redirect(url_for("login"))
+
+    @bp.route("/update-brand-app-name", methods=["POST"])
+    @authentication_required
```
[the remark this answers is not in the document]
rmlibre: Addressed by (https://github.com/scidsg/hushline/pull/573/commits/d847a509ef134cc86cea4dc00a4778d97e73d579)

hushline/settings.py
```diff
@@ -704,6 +712,50 @@ def update_smtp_settings() -> Response | str:
...
+    def update_brand_app_name() -> Response | str:
+        user_id = session.get("user_id")
+        if not user_id:
+            flash("Please log in to continue.")
+            return redirect(url_for("login"))
```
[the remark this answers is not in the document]
rmlibre: Addressed by (https://github.com/scidsg/hushline/pull/573/commits/d847a509ef134cc86cea4dc00a4778d97e73d579)

hushline/settings.py
```diff
@@ -704,6 +712,50 @@ def update_smtp_settings() -> Response | str:
         flash("👍 SMTP settings updated successfully")
         return redirect(url_for(".index"))
 
+    @bp.route("/update-brand-primary-color", methods=["POST"])
+    @authentication_required
+    def update_brand_primary_color() -> Response | str:
+        user_id = session.get("user_id")
+        if not user_id:
+            flash("Please log in to continue.")
+            return redirect(url_for("login"))
```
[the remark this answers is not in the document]
rmlibre: Addressed by (https://github.com/scidsg/hushline/pull/573/commits/d847a509ef134cc86cea4dc00a4778d97e73d579)

hushline/settings.py
```diff
@@ -704,6 +712,50 @@ def update_smtp_settings() -> Response | str:
...
+
+        user = db.session.get(User, user_id)
+        if user.is_admin:
+            # TODO:
+            # db persistence logic + form retrieval + update root style variable
+            form = UpdateBrandPrimaryColorForm()
+            if form.validate_on_submit():
+                user.hex_color = form.hex_color
```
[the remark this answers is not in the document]
rmlibre: Addressed by (https://github.com/scidsg/hushline/pull/573/commits/241b1df73bb09391f6ec0c635be905367aedd594)

hushline/__init__.py
```diff
@@ -61,6 +61,9 @@ def create_app() -> Flask:
     db.init_app(app)
     Migrate(app, db)
 
+    with app.app_context():
+        db.create_all()
```
jeremywmoore, line 65: is this needed? Don't we create the tables already?

hushline/model.py
```diff
@@ -28,6 +28,14 @@ def default(cls) -> "SMTPEncryption":
         return cls.StartTLS
 
 
+class HostOrganization(Model):
+    __tablename__ = "host_organization"
+
+    id: Mapped[int] = mapped_column(primary_key=True, default=1)
+    brand_app_name: Mapped[str] = mapped_column(db.String(255), default="🤫 Hush Line")
+    brand_primary_hex_color: Mapped[str] = mapped_column(db.String(255), default="#7d25c1")
```
jeremywmoore, line 36: should we limit the length to 7?

hushline/settings.py
```diff
@@ -395,6 +408,8 @@ async def index() -> str | Response:
             user=user,
             secondary_usernames=secondary_usernames,
             all_users=all_users,  # Pass to the template for admin view
+            update_brand_primary_color_form=update_brand_primary_color_form,
```
jeremywmoore, line 411: should we pre-populate the fields with the existing db values before this return statement?

rmlibre: I'm not sure. Because the tabs where the related forms are rendered require custom & coordinated attributes, I've manually defined them instead of relying completely on jinja inserts.


https://github.com/scidsg/hushline/blob/47d49da26ac78d123b0ad310de2b76f70ce2f222/hushline/templates/settings.html#L499-L507


https://github.com/scidsg/hushline/blob/47d49da26ac78d123b0ad310de2b76f70ce2f222/hushline/templates/settings.html#L516-L525


Pulling all the values from the forms & creating the whole tag + attributes using jinja could work, but that's not the existing pattern in this file. Additionally, accessing the ``host_org`` variable here

https://github.com/scidsg/hushline/blob/47d49da26ac78d123b0ad310de2b76f70ce2f222/hushline/settings.py#L286-L287

is a complication. I started with a version where ``host_org`` is accessed here, which caused db & key errors until the ``create_all()`` line in '\_\_init\_\_.py' was added https://github.com/scidsg/hushline./pull/573#discussion_r1761586224. But the ``create_all()`` line doesn't seem to be necessary when ``host_org`` is only accessed by the jinja templates.

I'm open to the idea that there's a good reason, or easy way, to rework how this is connected, but I don't think a change will be trivial.

rmlibre:
@jeremywmoore I'm unable to run prettier atm. It requires a version of node unavailable on my machine's default repositories. So I can't see the fixes needed to pass its linting tests.

hushline/__init__.py
rmlibre: Addressed by (https://github.com/scidsg/hushline/pull/573/commits/e9ca02df0fad9c3aa1e25102f599a0520d0340d7). Commented on also in (https://github.com/scidsg/hushline/pull/573#discussion_r1761779499).

hushline/model.py
rmlibre: Addressed by (https://github.com/scidsg/hushline/pull/573/commits/ab0a21251f9a94858515c724e86c00f3f172ec27)

glenn-sorrentino:
Looks great, @rmlibre. Let's make sure it's up-to-date with the latest and then I'm in favor of merging. @jeremywmoore thoughts?

hushline/forms.py
```diff
@@ -25,5 +26,22 @@ def __call__(self, form: Form, field: Field) -> None:
             raise ValidationError(self.message)
 
 
+class HexColor:
+    # HTML input color elements only give & accept 6-hexit color codes
+    hex_color_regex: re.Pattern = re.compile(r"^#[0-9a-fA-F]{6}$")
```
brassy-endomorph, line 31: Possibly want to make the `#` to `#?` to make it optional. Or the error should include the `#` so that users know what's missing

hushline/model.py
```diff
@@ -28,6 +28,14 @@ def default(cls) -> "SMTPEncryption":
         return cls.StartTLS
 
 
+class HostOrganization(Model):
+    __tablename__ = "host_organization"
+
+    id: Mapped[int] = mapped_column(primary_key=True, default=1)
```
brassy-endomorph, line 34: Primary keys shouldn't have a default. Also if we're expecting some "main org" to have a certain key, then we should make that a constant within the class so we can reference it elsewhere.

Also the two other fields shouldn't have defaults either, I don't think. The default should be at the class level or in templates otherwise changing the defaults will require Alembic migrations.

hushline/settings.py
```diff
@@ -704,6 +719,38 @@ def update_smtp_settings() -> Response | str:
         flash("👍 SMTP settings updated successfully")
         return redirect(url_for(".index"))
 
+    @bp.route("/update-brand-primary-color", methods=["POST"])
+    @admin_authentication_required
+    def update_brand_primary_color() -> Response | str:
+        if (host_org := db.session.get(HostOrganization, 1)) is None:
+            abort(500)
```
brassy-endomorph, line 726: Log an error here so an admin can figure out what went wrong.

hushline/settings.py
```diff
@@ -704,6 +719,38 @@ def update_smtp_settings() -> Response | str:
         flash("👍 SMTP settings updated successfully")
         return redirect(url_for(".index"))
 
+    @bp.route("/update-brand-primary-color", methods=["POST"])
+    @admin_authentication_required
+    def update_brand_primary_color() -> Response | str:
+        if (host_org := db.session.get(HostOrganization, 1)) is None:
```
brassy-endomorph, line 725: Use of a magic number. Should be a constant in the `HostOrganization` object.

hushline/__init__.py
```diff
@@ -77,6 +77,14 @@ def inject_user() -> dict[str, Any]:
             return {"user": user}
         return {}
 
+    @app.context_processor
+    def inject_host() -> dict[str, HostOrganization]:
+        if (host_org := db.session.get(HostOrganization, 1)) is None:
```
brassy-endomorph, line 82: Altering the DB in a context processor feels like an abuse of this feature. Adding the row should happen in the migration, and optionally we could have some CLI command called like `db-doctor` that ensures any required state is present in the DB so that admins have a single command they could run if somehow the DB gets into an unexpected state.

brassy-endomorph (commented):
Is it supposed to be the case that there will ever only be one row of `HostOrganization`? Because if so, we might want a generic table of `Setting`s that's just a key/value mapping storing values as `JSONB` so that we can more easily add others in the future without migrations.

hushline/forms.py
rmlibre: This is designed to work with the ``<input type="color">`` HTML element. It only accepts & produces exactly 6-hexit color codes with the hashtag.

Addressed by (https://github.com/scidsg/hushline/pull/573/commits/2cf9518c8cac279d9f03b1ee1b4b9bf8e5148295)

hushline/__init__.py
jeremywmoore: Instead of creating the `HostOrganization` here, can we add it via migration? Migrations aren't limited to DDL operations.

hushline/settings.py
rmlibre: Addressed by (https://github.com/scidsg/hushline/pull/573/commits/007cf1b1688684d8cf8751d28d8f4537788fb390)

hushline/model.py
rmlibre: The primary key was given a default so the application logic wouldn't be dependent on an implementation detail of how a given database system allocates primary keys (start at 0 or 1 or other?, randomized or incrementing?). I'd rather that defaults lower such cognitive burdens. However, I've moved the intended behavior to a new ``__init__()`` to avoid the strangeness of a default primary key.

There's only so much DRY that can be achieved with ease or without problems. For instance, the CSS hard encodes the default primary color here

https://github.com/scidsg/hushline/blob/2cf9518c8cac279d9f03b1ee1b4b9bf8e5148295/hushline/static/css/style.css#L5

which shouldn't be removed, because it's an independent file that should be self-contained & shouldn't break if not connected with the rest of the app infrastructure.

Instead of having the styles produce nulls as a measure to detect incoherent states, I'd rather there be actual runtime (or even better compile time) errors. Partially with that in mind, I've structured this table so that the whole organization is represented by a single record. It either exists or doesn't. If it doesn't, then it is initialized to a coherent state. If it is, then only validated values are allowed to change the already coherent state, which should keep it in a coherent state. By making each attribute a row, it becomes a cognitive burden to ensure when accessing a record that all the other associated records are also defined. It's valuable to remove potentials for inconsistency.

Addressed by (https://github.com/scidsg/hushline/pull/573/commits/007cf1b1688684d8cf8751d28d8f4537788fb390)

hushline/settings.py
rmlibre: Addressed by (https://github.com/scidsg/hushline/pull/573/commits/4097ffe5741400d260a55d27607f60e495481419)

hushline/__init__.py
brassy-endomorph: that's fine, but can you open a follow up for this once it's merged so it doesn't get lost?

hushline/model.py
```diff
@@ -31,9 +31,23 @@ def default(cls) -> "SMTPEncryption":
...
+    brand_app_name: Mapped[str] = mapped_column(db.String(255), default=_DEFAULT_BRAND_APP_NAME)
+    brand_primary_hex_color: Mapped[str] = mapped_column(
+        db.String(7), default=_DEFAULT_BRAND_PRIMARY_HEX_COLOR
+    )
+
+    @classmethod
+    def default(cls) -> Self | None:
+        return db.session.get(cls, cls._DEFAULT_PRIMARY_KEY)
+
+    def __init__(self, primary_key: int | None = None) -> None:
```
jeremywmoore, line 48: What's the purpose of manipulating the primary key in this way?

rmlibre: The overall intent is to be sure on record creation of the primary key value which is used for created records. I see depending on the implementation detail of how the database automatically assigns primary keys as a coupling which is unnecessary. The [review](https://github.com/scidsg/hushline/pull/573#discussion_r1764484915) by @brassy-endomorph also requested that the primary key not directly be given a default value in the ``mapped_column`` function. So, this initializer is intended meet that functionality while adhering to the request.

jeremywmoore: I don't see where the parameter is provided, should it be specified here, https://github.com/scidsg/hushline/pull/573/files#diff-567f5ecf57aaf3736dae09a470c91753792589003ee6b4ec067651aee9669afbR83

jeremywmoore: Ah, my bad, I was misreading this, I'm fine with this.

jeremywmoore:
Needs a rebase but otherwise I'm good with this PR

glenn-sorrentino:
@brassy-endomorph this may need your help getting merged.

hushline/settings.py
```diff
@@ -704,6 +719,40 @@ def update_smtp_settings() -> Response | str:
         flash("👍 SMTP settings updated successfully")
         return redirect(url_for(".index"))
 
+    @bp.route("/update-brand-primary-color", methods=["POST"])
+    @admin_authentication_required
+    def update_brand_primary_color() -> Response | str:
+        if (host_org := HostOrganization.default()) is None:
```
brassy-endomorph, line 725: I think we can just use `.one()` here and let the top level error handler report the full error and return 500 to the user. That tidies up the endpoint logic some, tho this is more a nit than a hard requirement.

hushline/__init__.py
```diff
@@ -77,6 +77,14 @@ def inject_user() -> dict[str, Any]:
             return {"user": user}
         return {}
 
+    @app.context_processor
+    def inject_host() -> dict[str, HostOrganization]:
```
brassy-endomorph, line 81: If the reason for doing an insert here is to ensure the app "just works" out of the box, maybe it makes more sense to do a `.one_or_none()` query and then return something like `HostOrganization.default()` that's maintained in pure python.

I'm still really iffy on this pattern as it feels like we're going to be populating the DB in too many places, and IMO the only places it should happen are:

- migrations (for prod)
- a CLI (for prod/dev)
- `dev_data.py` for dev

If we start scattering this sort of stuff throughout the app, I'm not sure about long term maintainability.

hushline/model.py
```diff
@@ -28,6 +28,28 @@ def default(cls) -> "SMTPEncryption":
...
 
+class HostOrganization(Model):
+    __tablename__ = "host_organization"
+
+    _DEFAULT_PRIMARY_KEY: int = 1
+    _DEFAULT_BRAND_PRIMARY_HEX_COLOR: str = "#7d25c1"
+    _DEFAULT_BRAND_APP_NAME: str = "🤫 Hush Line"
+
+    id: Mapped[int] = mapped_column(primary_key=True)
+    brand_app_name: Mapped[str] = mapped_column(db.String(255), default=_DEFAULT_BRAND_APP_NAME)
```
brassy-endomorph, line 39: I'm not sure how Alembic interprets `default` args to `mapped_column`, but we might need to make sure that the generated migrations don't have a server side default on the column. I think this happened to me earlier when testing some stuff on a different branch :shrug:

brassy-endomorph (approved):
tentative approval

hushline/model.py
jeremywmoore: I think `default=` sets the default value when the DB returns NULL. `server_default=` configures a default value in the table definition to be returned by the DB (https://docs.sqlalchemy.org/en/20/core/metadata.html#sqlalchemy.schema.Column.params.server_default)

hushline/__init__.py
brassy-endomorph: Thinking about this again that maybe a happy solution would be to inject a default value to the context but to _not_ add it to `db.session` and commit it. We should maybe have a global var and warn once (or maybe do this in `app.before_first_request` or something just so admins can see that maybe something is fucky if they don't have a row in the DB.

brassy-endomorph:
I'm trying to test this and at this commit (06042b7) with a clean dev env build, I'm getting this:

Screenshot 2024-09-25 at 18-38-58 Hush Line

A seemingly uncolored app. I'm not sure what the cause of this is since the CSS is too magical for my understanding. Any tips here?

brassy-endomorph:
@glenn-sorrentino or @rmlibre I need an assist here as I can't figure out why the app is unstyled on this branch or my own where I squashed/rebased these commits on top of `main`. https://github.com/scidsg/hushline/tree/customer-branding

Issue #586, opened 2024-09-14
SQLAlchemy metadata is acutally added to the extension


brassy-endomorph:
**Is your feature request related to a problem? Please describe.**

We create a naming convention for indices, but we don't add it to the `SQLAlchemy` extension

https://github.com/scidsg/hushline/blob/53c65eee7bf80365b6889fcb51cda268c68acd0e/hushline/db.py#L4-L14

**Describe the solution you'd like**

1. Add it
2. Regenerate all the migrations so that existing indices pick up the new naming convention

PR #593, opened 2024-09-18
Stripe CSS

glenn-sorrentino:
**Settings Banner:**
Screenshot 2024-09-17 at 5 10 50 PM

**Premium Page:**
Screenshot 2024-09-17 at 5 10 46 PM


migrations/versions/5c3fbcc6f32a_merging_two_heads.py
brassy-endomorph, line 1: It would be slightly better to just manually update the other revision files to create a linear history instead of adding another one of these "merge head" revisions.

migrations/versions/5c4788eea43e_add_stripe_events.py
brassy-endomorph, line 1: Could this be merged into the other revision so that there's just a singular stripe revision?

migrations/versions/677fc6ba8fba_create_tiers_table_and_add_default_tiers.py
brassy-endomorph, line 1: Likewise, merge revisions to have just one.

migrations/versions/86f4f7162aa0_add_stripe_fields.py
brassy-endomorph, line 1: merge revisions

migrations/versions/86f4f7162aa0_add_stripe_fields.py
```diff
@@ -0,0 +1,49 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "86f4f7162aa0"
+down_revision = "677fc6ba8fba"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
```
brassy-endomorph, line 21: nit: remove the comments of `commands auto generated by Alembic`

migrations/versions/88afd1ee90e8_add_status_to_stripeinvoice.py
brassy-endomorph, line 1: merge revisions

tests/test_premium.py
brassy-endomorph, line 1: empty file, remove

brassy-endomorph (commented):
Misc. other notes: 

- since we're committed to PG, should some of these primary keys be UUIDs to hide activity from other users?
- it would be good to preemptively name the indices in alignment with the metadata here so that we have consistent naming going forward and don't have to add a cleanup migration. #586
- i think it would be a generally good practice for single features to have single DB revisions so they can be wholly rolled out / back together

micahflee (changes requested):
Nice this looks great and could be merged now. But here are a few additional things I think could be helpful to focus on too.

When entering a credit card in dark mode, the font is black on dark grey:

image

After you submit credit card data it just says "Processing payment, please wait...". Should it stay like that? Should it show a cute graphic or animation?

image

After you pay, the Downgrade button looks exactly the same as the Upgrade button looked before. It might good to de-emphasize it? Like, keep it the same size, but just make it have a primary color border instead of background?

Also, can you stylize the invoice table? FYI the different possible invoice statuses are: "draft", "open", "paid", "uncollectible", and "void". In case you wanted to use icons or different colors or whatever to represent the status.

image

micahflee:
Actually, I'm just closing this PR. Please create a new one @glenn-sorrentino with my suggestions

PR #534, opened 2024-09-03
Paid features

micahflee:
This will address https://github.com/scidsg/hushline/issues/364

## How to test this

In one terminal, run the database:

```sh
./scripts/local-postgres.sh
```

In another terminal, forward the webhooks to the server using [stripe-cli](https://docs.stripe.com/stripe-cli) (you have to configure it first):

```sh
stripe listen --forward-to localhost:8080/premium/webhook
```

This will tell you something like:

> Your webhook signing secret is whsec_....

In another, set STRIPE_SECRET_KEY (you can find this in the Stripe dashboard at https://dashboard.stripe.com/test/apikeys) and STRIPE_WEBHOOK_SECRET (the stripe-cli command showed you that):

```sh
. dev_env.sh
export STRIPE_SECRET_KEY=sk_test_...
export STRIPE_WEBHOOK_SECRET=whsec_...
```

Run the migrations and initialize Stripe products:

```sh
make migrate
poetry run flask custom stripe
```

Then start server:

```sh
make run
```

With all of this running, when Stripe sends webhooks, it will make it to the local dev server.

===

Check out a demo of what it looks like so far. Clearly there's a lot of work to do in the design department, but the basic functionality works!

[Screencast from 2024-09-16 20-40-25.webm](https://github.com/user-attachments/assets/3ce5b254-a9f1-4791-9c6d-cd139fe118b3)

==

- [ ] On creating an account, allow the user to select free vs business. It can work like this: User.tier_id can be nullable. If it's null, on login, it prompts you to choose a tier.

Issue #364, opened 2024-06-07
MVP Stripe Integration


glenn-sorrentino:
This work has been started but isn't complete. You'll see a PR with a decent amount of progress, but it needs more attention. Feel free to use what makes sense or scrap it if not. 

## Acceptance Criteria

1. Basic paid tier.
2. Users can subscribe to unlock the secondary username feature.
3. Users should be able to cancel their subscription and retain paid features until their subscription period is complete.
4. Going to the "Advanced" tab shows a cancel subscription button.

## Design
![Group 314](https://github.com/user-attachments/assets/9109cc96-d1f1-4bdd-8cd2-96660115f8f9)

PR #595, opened 2024-09-18
Stripe integration

micahflee:
I'm closing the #534 PR and using this one instead. I've also merged #593 into it. Here's the instructions for testing again:

This will address https://github.com/scidsg/hushline/issues/364

## How to test this

In one terminal, run the database:

```sh
./scripts/local-postgres.sh
```

In another terminal, forward the webhooks to the server using [stripe-cli](https://docs.stripe.com/stripe-cli) (you have to configure it first):

```sh
stripe listen --forward-to localhost:8080/premium/webhook
```

This will tell you something like:

> Your webhook signing secret is whsec_....

In another, set STRIPE_SECRET_KEY (you can find this in the Stripe dashboard at https://dashboard.stripe.com/test/apikeys) and STRIPE_WEBHOOK_SECRET (the stripe-cli command showed you that):

```sh
. dev_env.sh
export STRIPE_SECRET_KEY=sk_test_...
export STRIPE_WEBHOOK_SECRET=whsec_...
```

Run the migrations and initialize Stripe products:

```sh
make migrate
poetry run flask custom stripe
```

Then start server:

```sh
make run
```

With all of this running, when Stripe sends webhooks, it will make it to the local dev server.


micahflee:
Things to do:

- [ ] When logging in, if user.tier_id is None, prompt the user to upgrade
- [ ] Show a link to the Premium page somewhere prominent (in the header? a tab in settings?)
- [ ] When a user downgrades, let them have the rest of their subscription term before downgrading (the [subscription object](https://docs.stripe.com/api/subscriptions/object) includes fields `current_period_end` and `current_period_start`, so it should be simple enough)

hushline/model.py
```diff
@@ -297,3 +307,106 @@ def __init__(self) -> None:
...
+
+class StripeEvent(Model):
+    __tablename__ = "stripe_events"
+
+    id: Mapped[int] = mapped_column(primary_key=True)
+    event_id: Mapped[str] = mapped_column(db.String(255), unique=True, index=True)
+    event_type: Mapped[str] = mapped_column(db.String(255))
+    event_data: Mapped[str] = mapped_column(db.Text)
+    created_at: Mapped[datetime] = mapped_column(default=datetime.now)
+    status: Mapped[str] = mapped_column(db.String(255), default="pending")
```
jeremywmoore, line 336: is there a defined set of fields here?

hushline/premium.py
```diff
@@ -0,0 +1,445 @@
...
+
+async def worker(app: Flask) -> None:
+    current_app.logger.error("Starting worker")
+    with app.app_context():
+        while True:
+            stripe_event = (
+                db.session.query(StripeEvent)
+                .filter_by(status="pending")
+                .order_by(StripeEvent.created_at)
+                .first()
```
jeremywmoore, line 235: I think this query should work. If there is a "pending" stripe event, it should update the status to "in_progress" and return the object, otherwise it returns `None`:
```suggestion
                db.session.scalars(db.update(StripeEvent)
                .where(status="pending")
                .order_by(StripeEvent.created_at)
                .values("in_progress")
                .limit(1)
                .returning(StripeEvent)).one_or_none()
```

hushline/premium.py
```diff
@@ -0,0 +1,445 @@
...
+                    if event.type == "invoice.created":
+                        handle_invoice_created(invoice)
+                    elif event.type == "invoice.payment_succeeded":
+                        handle_invoice_payment_succeeded(invoice)
+
+            except Exception as e:
+                current_app.logger.error(
+                    f"Error processing event {stripe_event.event_type} ({stripe_event.event_id}): {e}\n{stripe_event.event_data}"  # noqa: E501
+                )
+                stripe_event.status = "error"
```
jeremywmoore, line 277: should we save more error detail with the event?

hushline/premium.py
```diff
@@ -0,0 +1,445 @@
...
+        if stripe_subscription and stripe_subscription["status"] == "incomplete":
+            flash("⚠️ Your subscription is incomplete. Please try again.", "warning")
+
+        # Load the user's invoices
+        invoices = (
+            db.session.query(StripeInvoice)
+            .filter_by(user_id=user.id)
+            .filter_by(status=StripeInvoiceStatusEnum.PAID)
+            .order_by(desc(StripeInvoice.created_at))
+            .all()
```
jeremywmoore, line 316:
```suggestion
            db.session.scalars(db.select(StripeInvoice)
            .filter_by(user_id=user.id)
            .filter_by(status=StripeInvoiceStatusEnum.PAID)
            .order_by(desc(StripeInvoice.created_at))
            .all()
```

hushline/premium.py
```diff
@@ -0,0 +1,445 @@
...
+            session.clear()
+            return redirect(url_for("login"))
+
+        # If the user is already on the business tier
+        if user.tier_id == BUSINESS_TIER:
+            flash("👍 You're already upgraded.")
+            return redirect(url_for("premium.index"))
+
+        # Select the business tier
+        business_tier = db.session.query(Tier).get(BUSINESS_TIER)
```
jeremywmoore, line 343:
```suggestion
        business_tier = db.session.get(Tier, BUSINESS_TIER)
```

hushline/premium.py
```diff
@@ -0,0 +1,445 @@
...
+            )
+        except ValueError as e:
+            current_app.logger.error(f"Invalid payload: {e}")
+            return jsonify(success=False), 400
+        except stripe._error.SignatureVerificationError as e:
+            current_app.logger.error(f"Error verifying webhook signature: {e}")
+            return jsonify(success=False), 400
+
+        # Have we seen this one before?
+        stripe_event = db.session.query(StripeEvent).filter_by(event_id=event.id).first()
```
jeremywmoore, line 432:
```suggestion
        stripe_event = db.session.scalars(db.select(StripeEvent).filter_by(event_id=event.id)).one_or_none()
```

hushline/settings/__init__.py
```diff
@@ -274,6 +274,16 @@ async def index() -> str | Response:
             )
             user_count = len(all_users)
 
+        # Load the business tier price
+        business_tier = db.session.query(Tier).filter_by(name="Business").first()
```
jeremywmoore, line 278:
```suggestion
        business_tier = db.session.scalars(db.select(Tier).filter_by(name="Business")).one_or_none()
```

scripts/dev_data.py
```diff
@@ -47,6 +48,28 @@ def main() -> None:
...
+        },
+        {
+            "name": "Business",
+            "monthly_amount": 2000,
+        },
+    ]
+    for data in tiers:
+        name = str(data["name"])  # type: ignore
+        monthly_amount = int(data["monthly_amount"])  # type: ignore
+        if not db.session.query(exists(Tier).where(Tier.name == data["name"])).scalar():
```
jeremywmoore, line 64:
```suggestion
        if not db.session.scalar(db.exists(Tier).where(Tier.name == data["name"]).select()):
```

hushline/model.py
```diff
@@ -297,3 +307,106 @@ def __init__(self) -> None:
...
+            self.amount_remaining = invoice.amount_remaining
+        else:
+            self.amount_remaining = 0
+        if invoice.status:
+            self.status = StripeInvoiceStatusEnum(invoice.status)
+        if invoice.created:
+            self.created_at = datetime.fromtimestamp(invoice.created, tz=timezone.utc)
+
+        # Look up the user by their customer ID
+        user = db.session.query(User).filter_by(stripe_customer_id=invoice.customer).first()
```
jeremywmoore, line 396:
```suggestion
        user = db.session.scalars(db.select(User).filter_by(stripe_customer_id=invoice.customer)).one_or_none()
```

hushline/model.py
```diff
@@ -297,3 +307,106 @@ def __init__(self) -> None:
...
+        if user:
+            self.user_id = user.id
+        else:
+            raise ValueError(f"Could not find user with customer ID {invoice.customer}")
+
+        # Look up the tier by the product_id
+        if invoice.lines.data[0].plan:
+            product_id = invoice.lines.data[0].plan.product
+
+            tier = db.session.query(Tier).filter_by(stripe_product_id=product_id).first()
```
jeremywmoore, line 406:
```suggestion
            tier = db.session.scalars(db.select(Tier).filter_by(stripe_product_id=product_id)).one_or_none()
```

hushline/premium.py
```diff
@@ -0,0 +1,445 @@
...
+
+
+def init_stripe() -> None:
+    stripe.api_key = current_app.config["STRIPE_SECRET_KEY"]
+
+
+def create_products_and_prices() -> None:
+    current_app.logger.info("Creating products and prices")
+    # Make sure the products and prices are created in Stripe
+    tiers = db.session.query(Tier).all()
```
jeremywmoore, line 36:
```suggestion
    tiers = db.session.scalars(db.select(Tier)).all()
```

hushline/premium.py
```diff
@@ -0,0 +1,445 @@
...
+def get_subscription(user: User) -> stripe.Subscription | None:
+    if user.stripe_subscription_id is None:
+        return None
+
+    return stripe.Subscription.retrieve(user.stripe_subscription_id)
+
+
+def handle_subscription_created(subscription: stripe.Subscription) -> None:
+    # Subscription is created, but not yet paid
+    user = db.session.query(User).filter_by(stripe_customer_id=subscription.customer).first()
```
jeremywmoore, line 171:
```suggestion
    user = db.session.scalars(db.select(User).filter_by(stripe_customer_id=subscription.customer)).one_or_none()
```

hushline/premium.py
```diff
@@ -0,0 +1,445 @@
...
+    if user:
+        user.stripe_subscription_id = subscription.id
+        db.session.commit()
+    else:
+        raise ValueError(f"Could not find user with customer ID {subscription.customer}")
+
+
+def handle_subscription_updated(subscription: stripe.Subscription) -> None:
+    # If subscription changes to cancel or unpaid, downgrade user
+    user = db.session.query(User).filter_by(stripe_subscription_id=subscription.id).first()
```
jeremywmoore, line 181:
```suggestion
    user = db.session.scalars(db.select(User).filter_by(stripe_subscription_id=subscription.id)).one_or_none()
```

hushline/premium.py
```diff
@@ -0,0 +1,445 @@
...
+        if subscription.status in ["canceled", "unpaid"]:
+            user.tier_id = FREE_TIER
+        db.session.commit()
+    else:
+        raise ValueError(f"Could not find user with subscription ID {subscription.id}")
+
+
+def handle_subscription_deleted(subscription: stripe.Subscription) -> None:
+    # If subscription is deleted, downgrade user
+    user = db.session.query(User).filter_by(stripe_subscription_id=subscription.id).first()
```
jeremywmoore, line 192:
```suggestion
    user = db.session.scalars(db.select(User).filter_by(stripe_subscription_id=subscription.id)).one_or_none()
```

hushline/premium.py
```diff
@@ -0,0 +1,445 @@
...
+    try:
+        new_invoice = StripeInvoice(invoice)
+        db.session.add(new_invoice)
+        db.session.commit()
+    except ValueError as e:
+        current_app.logger.error(f"Error creating invoice: {e}")
+
+
+def handle_invoice_payment_succeeded(invoice: stripe.Invoice) -> None:
+    stripe_invoice = db.session.query(StripeInvoice).filter_by(invoice_id=invoice.id).first()
```
jeremywmoore, line 211:
```suggestion
    stripe_invoice = db.session.scalars(db.select(StripeInvoice).filter_by(invoice_id=invoice.id)).one_or_none()
```

hushline/settings/__init__.py
jeremywmoore: I noticed this seems to be the only place where we lookup the business tier by name instead of ID. Is that intentional?

hushline/premium.py
```diff
@@ -0,0 +1,445 @@
...
+    if stripe_invoice:
+        stripe_invoice.amount_paid = invoice.amount_paid
+        stripe_invoice.amount_remaining = invoice.amount_remaining
+        stripe_invoice.status = StripeInvoiceStatusEnum(invoice.status)
+        db.session.commit()
+
+        # Invoice has been paid, so make sure the user is upgraded
+        user = db.session.query(User).filter_by(stripe_subscription_id=invoice.subscription).first()
```
jeremywmoore, line 219:
```suggestion
        user = db.session.scalars(db.select(User).filter_by(stripe_subscription_id=invoice.subscription)).one_or_none()
```

jeremywmoore (commented):
Reviewed db queries (mostly for style)

hushline/model.py
```diff
@@ -297,3 +307,106 @@ def __init__(self) -> None:
...
+
+# Paid tiers
+class Tier(Model):
+    __tablename__ = "tiers"
+
+    id: Mapped[int] = mapped_column(primary_key=True)
+    name: Mapped[str] = mapped_column(db.String(255), unique=True)
+    monthly_amount: Mapped[int] = mapped_column(db.Integer)  # in cents USD
+    stripe_product_id = mapped_column(db.String(255), unique=True)
+    stripe_price_id = mapped_column(db.String(255), unique=True)
```
jeremywmoore, line 320: Needs type annotations
```suggestion
    stripe_product_id: Mapped[Optional[str]] = mapped_column(db.String(255), unique=True)
    stripe_price_id: Mapped[Optional[str]] = mapped_column(db.String(255), unique=True)
```
Based on the usage and alembic migration it looks like these are optional.

hushline/model.py
```diff
@@ -297,3 +307,106 @@ def __init__(self) -> None:
...
+
+    def __init__(self, name: str, monthly_amount: int) -> None:
+        super().__init__()
+        self.name = name
+        self.monthly_amount = monthly_amount
```
jeremywmoore, line 325: It looks like we're defining a constructor to require `name` and `monthly` amount which not allowing other columns to be specified on creation. In that case, we can still make use of the existing constructor:
```suggestion
        super().__init__(name=name, monthly_amount=monthly_amount)
```

hushline/model.py
```diff
@@ -297,3 +307,106 @@ def __init__(self) -> None:
...
+    event_type: Mapped[str] = mapped_column(db.String(255))
+    event_data: Mapped[str] = mapped_column(db.Text)
+    created_at: Mapped[datetime] = mapped_column(default=datetime.now)
+    status: Mapped[str] = mapped_column(db.String(255), default="pending")
+
+    def __init__(self, event: Event) -> None:
+        super().__init__()
+        self.event_id = event.id
+        self.event_type = event.type
+        self.event_data = str(event)
```
jeremywmoore, line 342:
```suggestion
        super().__init__(event_id=event.id, event_type=event.type, event_data=str(event))
```

hushline/model.py
```diff
@@ -297,3 +307,106 @@ def __init__(self) -> None:
...
+    VOID = "void"
+
+
+class StripeInvoice(Model):
+    __tablename__ = "stripe_invoices"
+
+    id: Mapped[int] = mapped_column(primary_key=True)
+    customer_id: Mapped[str] = mapped_column(db.String(255))
+    invoice_id: Mapped[str] = mapped_column(db.String(255), unique=True, index=True)
+    hosted_invoice_url: Mapped[str] = mapped_column(db.String(255))
```
jeremywmoore, line 359: How long are these URLs, should we up the limit to 2048 (the practical URL limit)?

hushline/model.py
```diff
@@ -297,3 +307,106 @@ def __init__(self) -> None:
...
+    amount_due: Mapped[int] = mapped_column(db.Integer)
+    amount_paid: Mapped[int] = mapped_column(db.Integer)
+    amount_remaining: Mapped[int] = mapped_column(db.Integer)
+    status: Mapped[StripeInvoiceStatusEnum] = mapped_column(
+        SQLAlchemyEnum(StripeInvoiceStatusEnum), nullable=True
+    )
```
jeremywmoore, line 365:
```suggestion
    status: Mapped[Optional[StripeInvoiceStatusEnum]] = mapped_column(
        SQLAlchemyEnum(StripeInvoiceStatusEnum)
    )
```

hushline/premium.py
```diff
@@ -0,0 +1,445 @@
...
+FREE_TIER = 1
+BUSINESS_TIER = 2
+
+
+def init_stripe() -> None:
+    stripe.api_key = current_app.config["STRIPE_SECRET_KEY"]
+
+
+def create_products_and_prices() -> None:
+    current_app.logger.info("Creating products and prices")
```
jeremywmoore, line 34: With multiple containers, do we need additional guards to make this safe for concurrency?

hushline/premium.py
```diff
@@ -0,0 +1,445 @@
...
+                db.session.add(stripe_event)
+                db.session.commit()
+                continue
+
+            stripe_event.status = "finished"
+            db.session.add(stripe_event)
+            db.session.commit()
+
+
+def start_event_loop(loop: asyncio.AbstractEventLoop) -> None:
```
jeremywmoore, line 287: I don't see a caller of this function

jeremywmoore (commented):
Another pass over model definitions

tests/test_premium.py
```diff
@@ -0,0 +1,422 @@
...
+        db.session.add(user)
+        db.session.commit()
+
+        subscription = get_subscription(user)
+
+        assert subscription is not None
+        assert mock_stripe.Subscription.retrieve.called
+
+
+def test_handle_subscription_created(app: Flask) -> None:
```
brassy-endomorph, line 175: Could you add a human explanation for this test so it can serve as documentation for our business logic?

tests/test_premium.py
```diff
@@ -0,0 +1,422 @@
...
+        subscription.current_period_end = datetime.now() + timedelta(days=30)
+        subscription.current_period_start = datetime.now()
+
+        handle_subscription_created(subscription)
+
+        assert user.stripe_subscription_id == "sub_123"
+        assert user.tier_id != BUSINESS_TIER
+
+
+def test_handle_subscription_updated_upgrade(app: Flask) -> None:
```
brassy-endomorph, line 196: Could you add a human explanation for this test so it can serve as documentation for our business logic?

tests/test_premium.py
```diff
@@ -0,0 +1,422 @@
...
+
+        subscription = get_subscription(user)
+
+        assert subscription is not None
+        assert mock_stripe.Subscription.retrieve.called
+
+
+def test_handle_subscription_created(app: Flask) -> None:
+    with app.app_context():
+        user = User(email="test@example.com", password="password")  # noqa: S106
```
brassy-endomorph, line 177: `user` fixture

tests/test_premium.py
```diff
@@ -0,0 +1,422 @@
...
+
+        handle_subscription_created(subscription)
+
+        assert user.stripe_subscription_id == "sub_123"
+        assert user.tier_id != BUSINESS_TIER
+
+
+def test_handle_subscription_updated_upgrade(app: Flask) -> None:
+    with app.app_context():
+        user = User(email="test@example.com", password="password")  # noqa: S106
```
brassy-endomorph, line 198: `user` fixture

tests/test_premium.py
```diff
@@ -0,0 +1,422 @@
...
+        subscription.current_period_start = datetime.now()
+
+        handle_subscription_updated(subscription)
+
+        assert user.tier_id == BUSINESS_TIER
+
+
+def test_handle_subscription_updated_downgrade(app: Flask) -> None:
+    with app.app_context():
+        user = User(email="test@example.com", password="password")  # noqa: S106
```
brassy-endomorph, line 217: `user` fixture

tests/test_premium.py
```diff
@@ -0,0 +1,422 @@
...
+        subscription.current_period_start = datetime.now()
+
+        handle_subscription_updated(subscription)
+
+        assert user.tier_id == FREE_TIER
+
+
+def test_handle_subscription_deleted(app: Flask) -> None:
+    with app.app_context():
+        user = User(email="test@example.com", password="password")  # noqa: S106
```
brassy-endomorph, line 236: `user` fixture

tests/test_premium.py
```diff
@@ -0,0 +1,422 @@
...
+
+        handle_subscription_deleted(subscription)
+
+        assert user.tier_id == FREE_TIER
+        assert user.stripe_subscription_id is None
+
+
+def test_handle_invoice_created(app: Flask) -> None:
+    with app.app_context():
+        user = User(email="test@example.com", password="password")  # noqa: S106
```
brassy-endomorph, line 252: `user` fixture

tests/test_premium.py
```diff
@@ -0,0 +1,422 @@
...
+                id="inv_123",
+                customer="cus_123",
+                hosted_invoice_url="https://example.com",
+                total=2000,
+                status=StripeInvoiceStatusEnum.OPEN,
+                lines=MagicMock(data=[MagicMock(plan=MagicMock(product="prod_123"))]),
+            )
+        )
+
+        stripe_invoice = db.session.query(StripeInvoice).filter_by(invoice_id="inv_123").first()
```
brassy-endomorph, line 268: `one()` and then assert has expected properties

tests/test_premium.py
```diff
@@ -0,0 +1,422 @@
...
+        assert stripe_invoice is not None
+
+
+def test_handle_invoice_updated(app: Flask) -> None:
+    with app.app_context():
+        user = User(email="test@example.com", password="password")  # noqa: S106
```
brassy-endomorph, line 274: `user` fixture

tests/test_premium.py
```diff
@@ -0,0 +1,422 @@
...
+            customer="cus_123",
+            hosted_invoice_url="https://example.com",
+            total=2000,
+            status=StripeInvoiceStatusEnum.OPEN,
+            lines=MagicMock(data=[MagicMock(plan=MagicMock(product="prod_123"))]),
+            subscription="sub_123",
+        )
+        handle_invoice_created(invoice)
+
+        stripe_invoice = db.session.query(StripeInvoice).filter_by(invoice_id="inv_123").first()
```
brassy-endomorph, line 291: `one()`

tests/test_premium.py
```diff
@@ -0,0 +1,422 @@
...
+        assert stripe_invoice is not None
+
+        invoice.status = StripeInvoiceStatusEnum.PAID
+        handle_invoice_updated(invoice)
+
+        stripe_invoice = db.session.query(StripeInvoice).filter_by(invoice_id="inv_123").first()
```
brassy-endomorph, line 297: `one()`

tests/test_premium.py
```diff
@@ -0,0 +1,422 @@
...
+) -> None:
+    user.tier_id = BUSINESS_TIER
+    db.session.commit()
+
+    response = client.post(url_for("premium.upgrade"))
+    assert response.status_code == 302
+    assert response.location == url_for("premium.index")
+
+
+# @pytest.mark.usefixtures("_authenticated_user")
```
brassy-endomorph, line 321: yank commented out code

migrations/versions/e3784181a957_add_stripe_tables.py
```diff
@@ -0,0 +1,157 @@
...
+        batch_op.drop_column("stripe_subscription_id")
+        batch_op.drop_column("stripe_customer_id")
+        batch_op.drop_column("tier_id")
+
+    with op.batch_alter_table("stripe_invoices", schema=None) as batch_op:
+        batch_op.drop_index(batch_op.f("ix_stripe_invoices_invoice_id"))
+
+    op.drop_table("stripe_invoices")
+    op.drop_table("tiers")
+    with op.batch_alter_table("stripe_events", schema=None) as batch_op:
```
brassy-endomorph, line 154: Don't need the explicit drop on the index since we drop the table in the next line.

hushline/templates/settings/profile.html
```diff
@@ -5,6 +5,20 @@
   aria-labelledby="profile-tab"
 >
   {# TODO: much of this was copy/pasted to "settings/alias.html" and will need to have updates synced there #}
+
+  {% if is_premium_enabled and (user.tier_id == 1 or user.tier_id == None) %}
```
brassy-endomorph, line 9: Nit. Move this message above the `TODO` so it's more obvious what the `TODO` references

hushline/templates/premium.html
```diff
@@ -0,0 +1,147 @@
+{% extends "base.html" %}
```
brassy-endomorph, line 1: Nit. For readability / skimability, newlines between the `{% ... %}` blocks

hushline/templates/premium-waiting.html
```diff
@@ -0,0 +1,11 @@
+{% extends "base.html" %}
```
brassy-endomorph, line 1: Nit. For readability / skimability, newlines between the `{% ... %}` blocks

hushline/settings/__init__.py
```diff
@@ -314,6 +324,9 @@ async def index() -> str | Response:
             pgp_key_percentage=pgp_key_percentage,
             directory_visibility_form=directory_visibility_form,
             default_forwarding_enabled=bool(current_app.config["NOTIFICATIONS_ADDRESS"]),
+            # Premium-specific data
+            is_premium_enabled=bool(current_app.config.get("STRIPE_SECRET_KEY", False)),
```
brassy-endomorph, line 328: This is a pattern we already use, but I wonder if somekinda `setup_g` pre-request function would be better for these types of derived configs. Or making a `stripe_ext` module with something like a `stripe.is_enabled` property or something. Don't need to change, just thinking aloud for future organization of the code.

hushline/premium.py
```diff
@@ -0,0 +1,536 @@
...
+                return jsonify(success=True)
+            except stripe._error.StripeError as e:
+                current_app.logger.error(f"Stripe error: {e}")
+                return jsonify(success=False), 400
+
+        return jsonify(success=False), 400
+
+    @bp.route("/status.json", methods=["GET"])
+    @authentication_required
+    def statis() -> Response | str:
```
brassy-endomorph, line 494: Typo. -> `status`

hushline/premium.py
```diff
@@ -0,0 +1,536 @@
...
+            return redirect(url_for("login"))
+
+        if user.tier_id == BUSINESS_TIER:
+            flash("🔥 Congratulations, you've upgraded your account!")
+
+        return jsonify({"tier_id": user.tier_id})
+
+    @bp.route("/webhook", methods=["POST"])
+    def webhook() -> Response | str | Tuple[Response | str, int]:
+        payload = request.data
```
brassy-endomorph, line 507: Unnecessary creation of local var that's used in only one place?

hushline/premium.py
```diff
@@ -0,0 +1,536 @@
...
+    def disable_autorenew() -> Response | str | Tuple[Response | str, int]:
+        user = db.session.get(User, session.get("user_id"))
+        if not user:
+            session.clear()
+            return redirect(url_for("login"))
+
+        if user.stripe_subscription_id:
+            try:
+                stripe.Subscription.modify(user.stripe_subscription_id, cancel_at_period_end=True)
+                user.stripe_subscription_cancel_at_period_end = True
```
brassy-endomorph, line 419: the `try/except` should only be around the `modify` call, the rest should come after or be in an `else` block

hushline/premium.py
```diff
@@ -0,0 +1,536 @@
...
+                    invoice: stripe.Invoice = stripe.Invoice.construct_from(
+                        event.data.object, current_app.config["STRIPE_SECRET_KEY"]
+                    )
+                    if event.type == "invoice.created":
+                        handle_invoice_created(invoice)
+                    elif event.type == "invoice.updated":
+                        handle_invoice_updated(invoice)
+
+            except Exception as e:
+                current_app.logger.error(
```
brassy-endomorph, line 309: We probably want the full stack trace and `exc_info=True` gets that for

hushline/premium.py
```diff
@@ -0,0 +1,536 @@
...
+        stripe_invoice.status = StripeInvoiceStatusEnum(invoice.status)
+        db.session.commit()
+    else:
+        raise ValueError(f"Could not find invoice with ID {invoice.id}")
+
+
+async def worker(app: Flask) -> None:
+    current_app.logger.error("Starting worker")
+    with app.app_context():
+        while True:
```
brassy-endomorph, line 265: Maybe not for this PR but as followup, do we want to set up some kinda signal handler / semaphore here instead of `while True` just so we can avoid possible errant state if this is killed in the middle of processing?

hushline/premium.py
```diff
@@ -0,0 +1,536 @@
...
+        user.stripe_subscription_current_period_start = None
+        db.session.commit()
+    else:
+        raise ValueError(f"Could not find user with subscription ID {subscription.id}")
+
+
+def handle_invoice_created(invoice: stripe.Invoice) -> None:
+    # invoice.created
+
+    try:
```
brassy-endomorph, line 242: Catch the `ValueError` only around the constructor?

brassy-endomorph (commented):
Lot of nits, few that are less nitty.

docker-compose.stripe.yaml
```diff
@@ -0,0 +1,56 @@
+---
```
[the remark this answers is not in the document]
micahflee: I spent a bit of time trying to figure out how to extend the default docker-compose file and couldn't get it to work properly... so I'll that for a future issue

Makefile
```diff
@@ -24,6 +24,11 @@ migrate-dev: ## Run dev env migrations
 migrate-prod: ## Run prod env (alembic) migrations
 	poetry run flask db upgrade
 
+.PHONY: dev-data
+dev-data: ## Run dev env migrations, and add dev data
+	poetry run ./scripts/dev_migrations.py && \
```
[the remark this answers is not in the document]
micahflee: Fixed in https://github.com/scidsg/hushline/pull/595/commits/0899cb5718744aa6e89f57832872ba992105c1f7

hushline/model.py
micahflee: Fixed in https://github.com/scidsg/hushline/pull/595/commits/c4594da873938a40623213fce086ee4b48601d2c

hushline/premium.py
micahflee: Fixed in https://github.com/scidsg/hushline/pull/595/commits/9c3af893d89787b5f716b42ddbe7849a8e1a276c

hushline/premium.py
micahflee: It now saved error messages too, in https://github.com/scidsg/hushline/pull/595/commits/217c78adaa08b5863f9cf699e2af415904f8c8b5

hushline/settings/__init__.py
micahflee: Nope that wasn't intentional. Fixed with https://github.com/scidsg/hushline/pull/595/commits/ebf6c998e3265785ed23bc94ca63326f198331e9

hushline/model.py
micahflee: They are pretty short, but who knows maybe someday they'll get longer so I'm good with 2048 chars

micahflee: Fixed in https://github.com/scidsg/hushline/pull/595/commits/c627d608ea751a5b26ce6795c2e9149f5e42defb

hushline/premium.py
micahflee: I've refactored this function so that it should be it's safe to run with multiple containers. When it runs it basically ensures  that the business tier's stripe_product_id and stripe_price_id match up with the Product and Price objects in Stripe. Running it multiple times shouldn't change anything, sort of like an ansible playbook.

docker-compose.stripe.yaml
```diff
@@ -0,0 +1,56 @@
...
+    environment:
+      FLASK_APP: hushline
+      FLASK_ENV: development
+      ENCRYPTION_KEY: bi5FDwhZGKfc4urLJ_ChGtIAaOPgxd3RDOhnvct10mw=
+      SECRET_KEY: cb3f4afde364bfb3956b97ca22ef4d2b593d9d980a4330686267cabcd2c0befd
+      SQLALCHEMY_DATABASE_URI: postgresql://hushline:hushline@postgres:5432/hushline
+      REGISTRATION_CODES_REQUIRED: False
+      SESSION_COOKIE_NAME: session
+      NOTIFICATIONS_ADDRESS: notifications@hushline.app
+    env_file:
```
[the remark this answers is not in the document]
micahflee: I'm using a separate env_file here in order to include secrets that shouldn't be committed to git, even in the dev environment (the test `STRIPE_SECRET_KEY`, which would give strangers the ability to do whatever they want in our test environment).

hushline/__init__.py
```diff
@@ -104,4 +120,32 @@ def add_onion_location_header(response: Response) -> Response:
             )
             return response
 
+    # Register custom CLI commands
+    register_commands(app)
+
     return app
+
+
+def register_commands(app: Flask) -> None:
```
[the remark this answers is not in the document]
micahflee: You mean, define these commands in the premium blueprint instead of in the app?

That's a good idea, except now I have refactored things so that `flask stripe configure` runs each time in dev and prod and ensures the default tiers (which are required for non-premium Hush Line instances too) are created. And this PR is getting gnarly, so I think I will put this off for a follow-up PR.

hushline/premium.py
```diff
@@ -0,0 +1,609 @@
...
+    User,
+)
+from .utils import authentication_required
+
+FREE_TIER = 1
+BUSINESS_TIER = 2
+
+
+def init_stripe() -> None:
+    stripe.api_key = current_app.config["STRIPE_SECRET_KEY"]
```
brassy-endomorph, line 40: Just noting that is pattern breaks Flask's app isolation and we're leaking configs between apps. This doesn't matter yet because of our current architecture, but just noting somehow so that if it ever causes problems at least one person other than me has been nudged toward thinking about this.

hushline/premium.py
```diff
@@ -0,0 +1,609 @@
...
+        db.session.commit()
+    else:
+        raise ValueError(f"Could not find invoice with ID {invoice.id}")
+
+
+async def worker(app: Flask) -> None:
+    current_app.logger.error("Starting worker")
+    with app.app_context():
+        while True:
+            stripe_event = db.session.scalars(
```
brassy-endomorph, line 303: This query/update/commit has a race condition because it's not done within a transaction block. If events can be processed multiple times, this isn't a problem.

hushline/premium.py
```diff
@@ -0,0 +1,609 @@
...
+        stripe_subscription = get_subscription(user)
+        if stripe_subscription and stripe_subscription["status"] == "incomplete":
+            flash("⚠️ Your subscription is incomplete. Please try again.", "warning")
+
+        # Load the user's invoices
+        invoices = db.session.scalars(
+            db.select(StripeInvoice)
+            .filter_by(user_id=user.id)
+            .filter_by(status=StripeInvoiceStatusEnum.PAID)
+            .order_by(desc(StripeInvoice.created_at))
```
brassy-endomorph, line 384: Nit. Can be `.order_by(StripeInvoice.created_at.desc())`

hushline/premium.py
```diff
@@ -0,0 +1,609 @@
...
+
+    @bp.route("/waiting", methods=["GET"])
+    @authentication_required
+    def waiting() -> Response | str:
+        return render_template("premium-waiting.html")
+
+    @bp.route("/upgrade", methods=["GET", "POST"])
+    @authentication_required
+    def upgrade() -> Response | str:
+        if request.method == "GET":
```
brassy-endomorph, line 427: Why are you redirecting on `GET`? Can we just make this a `POST` only endpoint?

hushline/premium.py
```diff
@@ -0,0 +1,609 @@
...
+        if not business_tier.stripe_price_id:
+            current_app.logger.error("Business tier does not have a price ID")
+            flash("⚠️ Something went wrong!")
+            return redirect(url_for("premium.index"))
+
+        # Make sure the user has a Stripe customer
+        try:
+            create_customer(user)
+        except stripe._error.StripeError as e:
+            current_app.logger.error(f"Failed to create Stripe customer: {e}")
```
brassy-endomorph, line 455: Might want `exc_info=True` in the logger to get more details instead of `{e}`

hushline/premium.py
```diff
@@ -0,0 +1,609 @@
...
+            checkout_session = stripe.checkout.Session.create(
+                client_reference_id=str(user.id),
+                customer=user.stripe_customer_id,
+                line_items=[{"price": business_tier.stripe_price_id, "quantity": 1}],
+                mode="subscription",
+                success_url=url_for("premium.waiting", _external=True),
+                automatic_tax={"enabled": True},
+                customer_update={"address": "auto"},
+            )
+        except stripe._error.StripeError as e:
```
brassy-endomorph, line 470: Might want `exc_info=True` in the logger to get more details instead of `{e}`

hushline/premium.py
```diff
@@ -0,0 +1,609 @@
...
+                db.session.add(user)
+                db.session.commit()
+
+                current_app.logger.info(
+                    f"Autorenew disabled for subscription {user.stripe_subscription_id} for user {user.id}"  # noqa: E501
+                )
+
+                flash("Autorenew has been disabled.")
+                return jsonify(success=True)
+            except stripe._error.StripeError as e:
```
brassy-endomorph, line 500: Might want `exc_info=True` in the logger to get more details instead of `{e}`

hushline/premium.py
```diff
@@ -0,0 +1,609 @@
...
+                db.session.add(user)
+                db.session.commit()
+
+                current_app.logger.info(
+                    f"Autorenew enabled for subscription {user.stripe_subscription_id} for user {user.id}"  # noqa: E501
+                )
+
+                flash("Autorenew has been enabled.")
+                return jsonify(success=True)
+            except stripe._error.StripeError as e:
```
brassy-endomorph, line 527: Might want `exc_info=True` in the logger to get more details instead of `{e}`

hushline/premium.py
```diff
@@ -0,0 +1,609 @@
...
+                db.session.add(user)
+                db.session.commit()
+
+                current_app.logger.info(
+                    f"Subscription {user.stripe_subscription_id} canceled for user {user.id}"
+                )
+
+                flash("💔 Sorry to see you go!")
+                return jsonify(success=True)
+            except stripe._error.StripeError as e:
```
brassy-endomorph, line 557: Might want `exc_info=True` in the logger to get more details instead of `{e}`

hushline/premium.py
```diff
@@ -0,0 +1,609 @@
...
+        payload = request.data
+        sig_header = request.headers["STRIPE_SIGNATURE"]
+
+        # Parse the event
+        try:
+            event = stripe.Webhook.construct_event(
+                payload, sig_header, current_app.config.get("STRIPE_WEBHOOK_SECRET")
+            )
+        except ValueError as e:
+            current_app.logger.error(f"Invalid payload: {e}")
```
brassy-endomorph, line 587: Slight possibility of logging sensitive data depending on what the `ValueError` contains. Not sure there's a way to log this. Maybe the webhook has an ID we can log and then reference it out of band?

hushline/templates/base.html
```diff
@@ -140,6 +140,9 @@ <h1>🤫 Hush Line</h1>
                     <li><a href="{{ url_for('logout') }}">Logout</a></li>
                   </ul>
                 </div>
+                {% if is_premium_enabled and (user.tier_id == 1 or user.tier_id == None) %}
```
brassy-endomorph, line 143: `is none` (lowercase) is a valid test in jinja

hushline/templates/premium-select-tier.html
```diff
@@ -0,0 +1,32 @@
...
+        action="{{ url_for('premium.upgrade') }}"
+        method="post"
+      >
+        <button>Use Business Plan</button>
+      </form>
+    </div>
+  </div>
+{% endblock %}
+
+{% block scripts %}
```
brassy-endomorph, line 31: Nit: If `scripts` are empty, the block can be omitted.

hushline/templates/premium.html
```diff
@@ -0,0 +1,96 @@
...
+{% block title %}Premium{% endblock %}
+{% block content %}
+  <h2>Choose a Hush Line Plan</h2>
+
+  <div id="plan-wrapper">
+    <div class="plan">
+      {% if user.tier_id == 1 or user.tier_id == None %}
```
brassy-endomorph, line 8: You have `user.tier_id == 1 or user.tier_id == None` a few times. It uses `1` as a magic number, so maybe the `User` should have an `is_free_tier` property?

hushline/templates/premium.html
```diff
@@ -0,0 +1,96 @@
...
+        <p class="plan-status badge">📍 Current Plan</p>
+      {% endif %}
+      {% if user.tier_id == 2 %}
```
brassy-endomorph, line 11: Magic number `2`. Maybe `is_business_tier` property for `User`?

tests/conftest.py
```diff
@@ -139,6 +154,7 @@ def user_password() -> str:
 @pytest.fixture()
 def user(app: Flask, user_password: str, database: str) -> User:
     user = User(password=user_password)
+    user.tier_id = 1
```
brassy-endomorph, line 157: Set this via the constructor?

hushline/model.py
```diff
@@ -144,6 +166,21 @@ class User(Model):
...
+
+    stripe_customer_id = mapped_column(db.String(255))
+    stripe_subscription_id = mapped_column(db.String(255), nullable=True)
+    stripe_subscription_cancel_at_period_end = mapped_column(
+        db.Boolean, default=False, nullable=True
+    )
+    stripe_subscription_status: Mapped[Optional[StripeSubscriptionStatusEnum]] = mapped_column(
+        SQLAlchemyEnum(StripeSubscriptionStatusEnum)
+    )
+    stripe_subscription_current_period_end = mapped_column(db.DateTime, nullable=True)
```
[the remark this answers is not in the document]
micahflee: Fixed in 0fe6e9ce705fcbd20554bd3f7bfcf12875edc510

hushline/model.py
```diff
@@ -297,3 +334,88 @@ def __init__(self) -> None:
...
+
+
+class StripeEvent(Model):
+    __tablename__ = "stripe_events"
+
+    id: Mapped[int] = mapped_column(primary_key=True)
+    event_id: Mapped[str] = mapped_column(db.String(255), unique=True, index=True)
+    event_type: Mapped[str] = mapped_column(db.String(255))
+    event_data: Mapped[str] = mapped_column(db.Text)
+    created_at: Mapped[datetime] = mapped_column(default=datetime.now)
```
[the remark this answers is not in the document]
micahflee: Fixed in bd7fbf27b034604e85e3b0d8116e4627ada832ef

hushline/model.py
```diff
@@ -297,3 +334,88 @@ def __init__(self) -> None:
...
+    total: Mapped[int] = mapped_column(db.Integer)
+    status: Mapped[StripeInvoiceStatusEnum] = mapped_column(
+        SQLAlchemyEnum(StripeInvoiceStatusEnum), nullable=True
+    )
+    created_at: Mapped[datetime] = mapped_column(default=datetime.now)
+
+    user_id: Mapped[int] = mapped_column(db.ForeignKey("users.id"))
+    tier_id: Mapped[int] = mapped_column(db.ForeignKey("tiers.id"))
+
+    def __init__(self, invoice: Invoice):
```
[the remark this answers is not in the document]
micahflee: Fixed in a50172dd4f8593a72a2ea670c0f55f297cdd640f

scripts/dev_data.py
```diff
@@ -47,6 +48,28 @@ def main() -> None:
...
+            "name": "Free",
+            "monthly_amount": 0,
+        },
+        {
+            "name": "Business",
+            "monthly_amount": 2000,
+        },
+    ]
+    for data in tiers:
+        name = str(data["name"])  # type: ignore
```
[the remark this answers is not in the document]
micahflee: Fixed in 100d2c45436a5ab4719a58e8a008cd74906c4fc5

tests/test_premium.py
```diff
@@ -0,0 +1,422 @@
...
+def business_tier() -> Tier:
+    tier = db.session.query(Tier).filter_by(name="Business").first()
+    if not tier:
+        raise ValueError("Business tier not found")
+
+    return tier
+
+
+def test_create_products_and_prices(app: Flask, mock_stripe: MagicMock) -> None:
+    with app.app_context():
```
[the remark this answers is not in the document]
micahflee: Fixed in 3b1c02e1841b6d536690fec338d580c0bfa5675c

tests/test_premium.py
```diff
@@ -0,0 +1,422 @@
...
+        # Remove the stripe_product_id and stripe_price_id from the Business tier
+        tier = Tier.query.filter_by(name="Business").first()
```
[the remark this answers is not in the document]
micahflee: Fixed in 44c4b8d3b19850248d565720096e1e0f3c1364bb

tests/test_premium.py
```diff
@@ -0,0 +1,422 @@
...
+        assert tier is not None
+        if not tier:
+            return
+
+        tier.stripe_product_id = None
```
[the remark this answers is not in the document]
micahflee: Commented

tests/test_premium.py
```diff
@@ -0,0 +1,422 @@
...
+
+        assert tier.stripe_product_id is not None
+        assert tier.stripe_price_id is not None
+
+
+def test_update_price_existing(app: Flask, mock_stripe: MagicMock) -> None:
+    mock_stripe.Price.search.return_value = [MagicMock(id="price_123", unit_amount=2000)]
+
+    with app.app_context():
+        tier = Tier.query.filter_by(name="Business").first()
```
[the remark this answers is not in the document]
micahflee: Fixed in 44c4b8d3b19850248d565720096e1e0f3c1364bb

tests/test_premium.py
```diff
@@ -0,0 +1,422 @@
...
+        db.session.refresh(user)
+        assert user.stripe_customer_id is not None
+
+
+def test_create_subscription(app: Flask, mock_stripe: MagicMock) -> None:
+    mock_stripe.Customer.create.return_value = MagicMock(id="cus_123")
+    mock_stripe.Subscription.create.return_value = MagicMock(id="sub_123")
+
+    with app.app_context():
+        tier = Tier.query.filter_by(name="Business").first()
```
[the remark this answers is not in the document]
micahflee: Fixed in 44c4b8d3b19850248d565720096e1e0f3c1364bb

tests/test_premium.py
micahflee: Fixed in dd6824ec698dfdda90ef67d3f5a60fb331ddcf1a

tests/test_premium.py
micahflee: Fixed in dd6824ec698dfdda90ef67d3f5a60fb331ddcf1a

tests/test_premium.py
```diff
@@ -0,0 +1,422 @@
...
+        assert tier is not None
+        if not tier:
+            return
+
+        update_price(tier)
+
+        assert mock_stripe.Price.create.called
+        assert mock_stripe.Product.modify.called
```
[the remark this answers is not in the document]
micahflee: Fixed in 5c62beebbadfded4d7818a59948dfd7a55ddeab3

tests/test_premium.py
```diff
@@ -0,0 +1,422 @@
...
+
+
+def test_create_customer(app: Flask, mock_stripe: MagicMock) -> None:
+    mock_stripe.Customer.create.return_value = MagicMock(id="cus_123")
+
+    with app.app_context():
+        user = User(email="test@example.com", password="password")  # noqa: S106
```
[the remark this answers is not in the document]
micahflee: Fixed in 180d5b3514b575c43e6b99320d8471a6f341796f

tests/test_premium.py
```diff
@@ -0,0 +1,422 @@
...
+
+def test_create_subscription(app: Flask, mock_stripe: MagicMock) -> None:
+    mock_stripe.Customer.create.return_value = MagicMock(id="cus_123")
+    mock_stripe.Subscription.create.return_value = MagicMock(id="sub_123")
+
+    with app.app_context():
+        tier = Tier.query.filter_by(name="Business").first()
+        assert tier is not None
+
+        user = User(email="test@example.com", password="password")  # noqa: S106
```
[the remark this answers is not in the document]
micahflee: Fixed in 180d5b3514b575c43e6b99320d8471a6f341796f

tests/test_premium.py
micahflee: Fixed in 180d5b3514b575c43e6b99320d8471a6f341796f

tests/test_premium.py
micahflee: Fixed in 180d5b3514b575c43e6b99320d8471a6f341796f

tests/test_premium.py
micahflee: Fixed in 180d5b3514b575c43e6b99320d8471a6f341796f

tests/test_premium.py
micahflee: Fixed in 180d5b3514b575c43e6b99320d8471a6f341796f

tests/test_premium.py
micahflee: Fixed in abf7cb230fb2a4417a50e6351ef11cb08182ffa3

tests/test_premium.py
micahflee: Fixed in abf7cb230fb2a4417a50e6351ef11cb08182ffa3

tests/test_premium.py
micahflee: Fixed

tests/test_premium.py
micahflee: Fixed

tests/test_premium.py
micahflee: Fixed in e359867183a8f1705b1fffe39f7c239118d3f609

migrations/versions/e3784181a957_add_stripe_tables.py
micahflee: Fixed in 8984e5765a1295c8a4f7fb4dfa038b4563ed43c7

hushline/templates/settings/profile.html
micahflee: Fixed in 28932912dd6fbb8ede5c3ecbfc05660285da3064

hushline/templates/premium-waiting.html
micahflee: Fixed in f4c3e88573128e2960908e0dac6ee4d44c46e331

hushline/settings/__init__.py
micahflee: I actually refactored this already to make `is_premium_enabled` a global config available in all templates

hushline/premium.py
micahflee: Fixed in 3cf0145b89536fb87de6b105a040d7ff33f99676

hushline/premium.py
micahflee: Fixed in ff5f2173ca30559812e5b4a9652b6355b061200a

hushline/premium.py
micahflee: Fixed in ff5f2173ca30559812e5b4a9652b6355b061200a

hushline/premium.py
micahflee: Fixed in ff5f2173ca30559812e5b4a9652b6355b061200a

hushline/premium.py
micahflee: Fixed in ff5f2173ca30559812e5b4a9652b6355b061200a

tests/conftest.py
```diff
@@ -115,13 +115,28 @@ def app(database: str) -> Generator[Flask, None, None]:
...
     app = create_app()
     app.config["TESTING"] = True
     app.config["WTF_CSRF_ENABLED"] = False
     app.config["SERVER_NAME"] = "localhost:8080"
     app.config["PREFERRED_URL_SCHEME"] = "http"
 
     with app.app_context():
+        db.create_all()
+
+        # Create the default tiers
```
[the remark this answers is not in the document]
micahflee: I tried to do this an kept running into issues... so let's save it for a later PR. But yes we need the tiers to be created each time, not just when testing premium stuff, as the templates and stuff depend on it.

hushline/premium.py
micahflee: Good idea for the future. I'm not sure the best way to handle it. Though I expect we will only actually want one worker container running at any time, since it's only used for Stripe webhook events. Hush Line would have to be getting a massive amounts of money for this to hit any scaling problems

hushline/premium.py
micahflee: Fixed in 438630f4a5efc22204768c950bd78e514ef530cd

hushline/premium.py
micahflee: Yeah, I don't know why I was doing that

hushline/premium.py
micahflee: Fixed in 977935a15167450c085c93340cb31e0ac6efca01

hushline/premium.py
micahflee: Fixed in 977935a15167450c085c93340cb31e0ac6efca01

hushline/premium.py
micahflee: Fixed in 977935a15167450c085c93340cb31e0ac6efca01

hushline/premium.py
micahflee: Fixed in 977935a15167450c085c93340cb31e0ac6efca01

hushline/premium.py
micahflee: Fixed in 977935a15167450c085c93340cb31e0ac6efca01

hushline/premium.py
micahflee: I'm not too concerned because the webhook events shouldn't actually contain sensitive data. No credit card numbers or anything.

hushline/premium.py
micahflee: Fixed in 01603c0d5b5b23eec26e23dae75c84dfc1fd839b

tests/conftest.py
micahflee: We actually want tier_id to start out as None by default, because on first login, if it's None, it displays the tier selection page. But in the tests, I wanted them all to start out on the free tier, to skip the selection page.

hushline/templates/base.html
micahflee: Fixed in d57e2adb3ce90717ed32aaf6fd96b5e1ec1c643b

hushline/templates/premium-select-tier.html
micahflee: Fixed in d57e2adb3ce90717ed32aaf6fd96b5e1ec1c643b

hushline/templates/premium.html
micahflee: Good idea. Fixed in 0e3d8557c9d60ffbef0a62872889841223831535

hushline/premium.py
brassy-endomorph: I suspected no either but wanted to flag it at least

tests/conftest.py
brassy-endomorph: Oh sorry I meant just `User(password=user_password, tier_id=1)`

micahflee:
This is just about ready for a final review, as well as for some design work! Let me show how it all works.

When new users register accounts, the first time they login they are prompted to select a tier:

image

This displays the `premium-select-tier.html` template, which includes subtemplates for the free and business feature lists that are shared with the `premium.html` template now. Also, in `premium/business-features.html`, where it says "Monthly Price: $20", it actually pulls the $20 from whatever we set the price to, it's no longer hard-coded. So if we change the price, it will change there too.

Once you select a plan (I selected free) it no longer displays that on login.

If you're logged into the free tier, there's now an "Upgrade" button in the nav bar. You may want to change how it looks:

image

Clicking it brings you to `/premium/`. It's the same as the link in the Profile page of Settings:

image

Here's the premium index page:

image

I've switched it to use Stripe Checkout. If you click "Upgrade Now", it now creates a new checkout session and redirects to a Stripe-hosted page to collect payment info:

image

After filling out the payment info, it briefly shows a waiting page (`premium-waiting.html`, though it went by too quickly to get a screenshot) and then loads the premium page again:

image

This time, now that the user is signed in and using a business account, it no longer has the Upgrade button in the nav bar. It also displays a table of invoices at the bottom (this part needs design work). And instead of just canceling the subscription, there's a whole workflow for disabling and re-enabling auto-renew. Here's a short video showing it:

[Screencast from 2024-10-01 10-21-14.webm](https://github.com/user-attachments/assets/2a6e1464-8ba3-4db9-9fea-e439fd6ecff4)

The buttons need some design work too, but it works!

If you're on a business plan, and you load the profile page of Settings, it should no longer show the Upgrade button there:

image

And if you go to the Advanced tab, it should now have a link to manage your plan (this definitely needs some design work):

image

That just links back to the premium page, where you can choose to not renew/cancel your subscription, or access your invoices.

===

All of these premium features should only happen if the `STRIPE_SECRET_KEY` environment variable is set. It injects `is_premium_enabled` into the templates in `hushline/__init__.py`:

```py
@app.context_processor
def inject_is_premium_enabled() -> dict[str, Any]:
    return {"is_premium_enabled": bool(app.config.get("STRIPE_SECRET_KEY", False))}
```

So an important thing to check for while reviewing is that if premium mode is disabled (no `STRIPE_SECRET_KEY`), that none of these changes appear.

This PR also adds a Tier model with two tiers, free and business. It adds a bunch of new fields to the User model:

- tier_id
- stripe_customer_id
- stripe_subscription_id
- stripe_subscription_cancel_at_period_end
- stripe_subscription_status
- stripe_subscription_current_period_end
- stripe_subscription_current_period_start

And it adds models for StripeInvoice and StripeEvent.

===

Stripe webhooks work like this:

The webhook URL is `/premium/webhook`, so in prod we will need to login to the Stripe dashboard and configure a webhook to use the URL https://tips.hushline.app/premium/webhook. This route is configured to verify the signature on the webhook event (to make sure it came from Stripe and not randos making POST requests to that URL), and then insert the event into the database (with a "pending" status) and return 200.

The premium features will require an extra worker container, so before we can deploy this to prod, there needs to be some infra work to spin up an extra container. The worker container is just the app contain but it starts with this command: `poetry run flask stripe start-worker`

The worker container runs an infinite loop looking for pending Stripe events and processing them. Right now it only handles the following events (which are all that's needed for the functionality we're using):

- customer.subscription.created
- customer.subscription.updated
- customer.subscription.deleted
- invoice.created
- invoice.updated

So when creating the webhook in the Stripe dashboard, we can limit it to these events.

hushline/model.py
```diff
@@ -75,6 +77,40 @@ class ExtraField:
...
+@enum.unique
+class StripeEventStatusEnum(enum.Enum):
+    PENDING = "pending"
+    IN_PROGRESS = "in_progress"
+    ERROR = "error"
+    FINISHED = "finished"
+
+
+# Premium tiers
+FREE_TIER = 1
```
brassy-endomorph, line 110: We might want to move these into the `Tier` model and stick to that pattern generally, where a constant related to a class is always part of that class.

hushline/model.py
```diff
@@ -188,6 +225,25 @@ class User(Model):
...
     smtp_sender: Mapped[Optional[str]]
 
+    # Paid tier fields
+    tier_id: Mapped[int | None] = mapped_column(db.ForeignKey("tiers.id"), nullable=True)
+    tier: Mapped["Tier"] = relationship(backref=db.backref("tiers", lazy=True))
+
+    stripe_customer_id = mapped_column(db.String(255))
+    stripe_subscription_id = mapped_column(db.String(255), nullable=True)
+    stripe_subscription_cancel_at_period_end = mapped_column(
+        db.Boolean, default=False, nullable=True
```
brassy-endomorph, line 235: We have `default=False` but also `nullable=True` which might mean we never have `null` value set. Do we ever need this to be `null`? Or can we force it to be always false? Also slight preference for `server_default=text('false')`.

hushline/model.py
```diff
@@ -349,3 +411,96 @@ def __init__(self) -> None:
...
+
+    id: Mapped[int] = mapped_column(primary_key=True)
+    name: Mapped[str] = mapped_column(db.String(255), unique=True)
+    monthly_amount: Mapped[int] = mapped_column(db.Integer)  # in cents USD
+    stripe_product_id: Mapped[Optional[str]] = mapped_column(db.String(255), unique=True)
+    stripe_price_id: Mapped[Optional[str]] = mapped_column(db.String(255), unique=True)
+
+    def __init__(self, name: str, monthly_amount: int) -> None:
+        super().__init__()
+        self.name = name
```
brassy-endomorph, line 428: Oneliner -> `super().__init__(name=name, monthly_amount=monthly_amount)`. Though `mypy` might yell about that. :thinking:

hushline/model.py
```diff
@@ -349,3 +411,96 @@ def __init__(self) -> None:
...
+    event_type: Mapped[str] = mapped_column(db.String(255))
+    event_created: Mapped[int] = mapped_column(db.Integer)
+    event_data: Mapped[str] = mapped_column(db.Text)
+    status: Mapped[Optional[StripeEventStatusEnum]] = mapped_column(
+        SQLAlchemyEnum(StripeEventStatusEnum), default=StripeEventStatusEnum.PENDING
+    )
+    error_message: Mapped[Optional[str]] = mapped_column(db.Text)
+
+    def __init__(self, event: Event) -> None:
+        super().__init__()
```
brassy-endomorph, line 446: Pass args as `kwargs` in `super().__init__(...)`

scripts/dev_data.py
```diff
@@ -47,6 +50,28 @@ def main() -> None:
...
+        },
+        {
+            "name": "Business",
+            "monthly_amount": 2000,
+        },
+    ]
+    for data in tiers:
+        name = cast(str, data["name"])
+        monthly_amount = cast(int, data["monthly_amount"])
+        if not db.session.scalar(db.exists(Tier).where(Tier.name == data["name"]).select()):
```
brassy-endomorph, line 66: `Tier.nname == name` since we're already defined that

migrations/versions/be0744a5679f_add_stripe_tables.py
```diff
@@ -0,0 +1,154 @@
...
+                nullable=True,
+            ),
+        )
+        batch_op.create_index("idx_users_stripe_customer_id", ["stripe_customer_id"], unique=False)
+        batch_op.create_foreign_key("users_tier_id_fkey", "tiers", ["tier_id"], ["id"])
+
+
+def downgrade() -> None:
+    with op.batch_alter_table("users", schema=None) as batch_op:
+        batch_op.drop_constraint("users_tier_id_fkey", type_="foreignkey")
```
brassy-endomorph, line 140: This doesn't match the pattern in the metadata which is `"fk_%(table_name)s_%(column_0_name)s_%(referred_table_name)s"`. Having it differ would mean other migrations would drop/create it.

hushline/templates/premium.html
```diff
@@ -0,0 +1,98 @@
...
+          id="upgrade-form"
+          action="{{ url_for('premium.upgrade') }}"
+          method="post"
+        >
+          <button id="upgrade">Upgrade Now</button>
+        </form>
+      {% endif %}
+      {% if user.is_business_tier() %}
+        {% if user.stripe_subscription_cancel_at_period_end %}
+        <p>Your subscription will expire on {{ user.stripe_subscription_current_period_end.strftime('%B %d, %Y') }}.</p>
```
brassy-endomorph, line 37: Nit: This and the next forms aren't indented relative to their enclosing `if/else` block

hushline/premium.py
```diff
@@ -0,0 +1,608 @@
...
+
+                flash("💔 Sorry to see you go!")
+                return jsonify(success=True)
+            except stripe._error.StripeError as e:
+                current_app.logger.error(f"Stripe error: {e}", exc_info=True)
+                return jsonify(success=False), 400
+
+        return jsonify(success=False), 400
+
+    @bp.route("/status.json", methods=["GET"])
```
brassy-endomorph, line 563: Nit: `methods` are unnecessary when the endpoint is only `GET`

hushline/premium.py
```diff
@@ -0,0 +1,608 @@
...
+    @authentication_required
+    def status() -> Response | str:
+        user = db.session.get(User, session.get("user_id"))
+        if not user:
+            session.clear()
+            return redirect(url_for("login"))
+
+        if user.tier_id == BUSINESS_TIER:
+            flash("🔥 Congratulations, you've upgraded your account!")
```
brassy-endomorph, line 572: `flash`-ing a message from a JSON endpoint feels wrong. What's the use case here?

hushline/premium.py
```diff
@@ -0,0 +1,608 @@
...
+                db.session.add(user)
+                db.session.commit()
+
+                current_app.logger.info(
+                    f"Autorenew enabled for subscription {user.stripe_subscription_id} for user {user.id}"  # noqa: E501
+                )
+
+                flash("Autorenew has been enabled.")
+                return jsonify(success=True)
+            except stripe._error.StripeError as e:
```
brassy-endomorph, line 527: Might want to tighten the `except` block to just be around the `Subscription.modify` call

hushline/premium.py
```diff
@@ -0,0 +1,608 @@
...
+    @bp.route("/upgrade", methods=["POST"])
+    @authentication_required
+    def upgrade() -> Response | str:
+        user = db.session.get(User, session.get("user_id"))
+        if not user:
+            session.clear()
+            return redirect(url_for("login"))
+
+        # If the user is already on the business tier
+        if user.tier_id == BUSINESS_TIER:
```
brassy-endomorph, line 433: Could use `is_business_tier` here

hushline/model.py
```diff
@@ -256,6 +312,12 @@ def pgp_key(self, value: str) -> None:
         else:
             self._pgp_key = encrypt_field(value)
 
+    def is_free_tier(self) -> bool:
```
brassy-endomorph, line 315: Make these `@property`s?

hushline/__init__.py
```diff
@@ -126,3 +146,43 @@ def add_onion_location_header(response: Response) -> Response:
...
+    stripe_cli = AppGroup("stripe")
+
+    @stripe_cli.command("configure")
+    def configure() -> None:
+        """Configure Stripe and premium tiers"""
+        # Make sure tiers exist
+        with app.app_context():
+            free_tier = db.session.query(Tier).filter_by(name="Free").first()
```
brassy-endomorph, line 159: We have several places where we use the strings `"Business"` and `"Free"` as well as their price. Perhaps these should be constants in the `Tier` model? Or perhaps since there's a few places create them, we have `@classmethod`s for `create_{free,business}_tier()`

hushline/__init__.py
```diff
@@ -126,3 +146,43 @@ def add_onion_location_header(response: Response) -> Response:
...
+            if not free_tier:
```
brassy-endomorph, line 160: Bigger philosophical question that came in the last PR. Do we want migrations to handle this sort of thing or app code? I feel like pre-populating tables should go in a migration so we're guaranteed that if the schema exists it has correct values.

brassy-endomorph (approved):
Went through the manual test plan. Everything works as advertised. I have some open remarks that might be good to address, but those can be in follows to. Approved.

hushline/model.py
micahflee: Fixed in 3757b5ec6137201111e00ed8335b73f8e5388721

hushline/premium.py
micahflee: When a user upgrades, they go to Stripe Checkout and fill out their credit card info, then get redirected to /premium/waiting. This page uses javascript to poll /premium/status (this JSON endpoint) every few seconds until the webhoooks successfully come in. When they do, it redirects to /premium and displays the "Congratulations" flash message.

This is the only use of this JSON endpoint, and there isn't really a better place to put the flash message since we don't know that the user has actually been upgraded until this point.

hushline/model.py
micahflee: Fixed in 3757b5ec6137201111e00ed8335b73f8e5388721

hushline/__init__.py
micahflee: Fixed in 3757b5ec6137201111e00ed8335b73f8e5388721

hushline/model.py
micahflee: mypy yells at me :(

hushline/__init__.py
micahflee: I will choose to ignore philosophy in the service of merging this PR sooner lol

hushline/model.py
micahflee: Fixed in 40b1ca102ea381b325f06f6a1ec5ceb61c4228a2

hushline/model.py
micahflee: Fixed in 40b1ca102ea381b325f06f6a1ec5ceb61c4228a2

migrations/versions/be0744a5679f_add_stripe_tables.py
micahflee: Fixed in 40b1ca102ea381b325f06f6a1ec5ceb61c4228a2

hushline/premium.py
micahflee: Fixed in 40b1ca102ea381b325f06f6a1ec5ceb61c4228a2

hushline/premium.py
micahflee: Fixed in 40b1ca102ea381b325f06f6a1ec5ceb61c4228a2

glenn-sorrentino (approved):

PR #599, opened 2024-09-19
Add app contents to Dockerfile.dev to enable use in remote environments

jeremywmoore:
This PR adds the contents of the hushline repo to Dockerfile.dev to facilitate running tools/scripts in remove dev environments where we don't have volume mounts.

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -22,3 +22,5 @@ ENV PYTHONPATH=/app
 COPY pyproject.toml poetry.lock .
 
 RUN poetry install
+
+COPY . /app


jeremywmoore:
Alternatively, we could update `Dockerfile.prod` to remove the `--no-root` option since we no longer need to optimize builds for local development.

brassy-endomorph:
I'm lost again on this PR. We mentioned some bug, but I can't find discussion of it. What was this trying to fix?

jeremywmoore:
@brassy-endomorph no worries, this PR is intended to support running the dev_data.py script in the context of remote dev deployments. The image built by Dockerfile.dev in its current form isn't suitable because it doesn't contain the project source (the script or hushline app source). The image built by Dockerfile.prod isn't suitable as it doesn't install hushline into the venv, only its dependency (--no-root).  The two options are 1) Add the source to the Dockerfile.dev or 2) install the application without the `--no-root` option. I chose option 1 here since adding test data feels like a "dev" type of activity and when we use the dev container locally, it has the source code mounted as a volume.

Alternatively, we could make the change presented in https://github.com/scidsg/hushline/pull/619 and run the dev_data script from an image build from `Dockerfile.prod`.

brassy-endomorph:
Both of these feel like kluges and if we're going to include `dev_data.py` in a prod container, we might as well just make it a normal `click` command and put it in the app so we can avoid all these oddities. Double `poetry install` on the prod container and copying the project root to `/app` and then also mounting the project root to `/app` in the compose file both feel like anti-patterns to me.

Should we just make it a command then? Can do the same with the `dev_migrations.py` too at that point

PR #611, opened 2024-09-20
run prettier via docker container

jeremywmoore:
This change runs prettier via docker container. It also defines the scope of prettier to `.md` files in the repo root, github actions workflows, docs, and the hushline directory.

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -6,11 +6,17 @@ RUN apt-get update && \
     apt-get install -y --no-install-recommends \
         build-essential \
         libffi-dev \
-        libpq-dev
+        libpq-dev \
+        npm
 
 RUN pip install poetry
 
 WORKDIR /app
+
+COPY package.json package-lock.json .
+
+RUN npm install
+
 ENV PYTHONPATH=/app
 
 COPY pyproject.toml poetry.lock .
--- a/Makefile
+++ b/Makefile
@@ -28,13 +28,13 @@ lint: ## Lint the code
 	poetry run ruff format --check && \
 	poetry run ruff check && \
 	poetry run mypy .
-	docker compose run --rm -v $(shell pwd):/work app npx prettier --check ./*.md ./docs ./.github/workflows/* ./hushline
+	docker compose run --rm app npx prettier --check ./*.md ./docs ./.github/workflows/* ./hushline
 
 .PHONY: fix
 fix: ## Format the code
 	poetry run ruff format && \
 	poetry run ruff check --fix
-	docker compose run --rm -v $(shell pwd):/work app npx prettier --write ./*.md ./docs ./.github/workflows/* ./hushline
+	docker compose run --rm app npx prettier --write ./*.md ./docs ./.github/workflows/* ./hushline
 
 .PHONY: revision
 revision: migrate-prod ## Create a new migration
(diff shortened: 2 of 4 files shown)


Dockerfile.dev
```diff
@@ -12,6 +12,6 @@ RUN pip install poetry
 WORKDIR /app
 ENV PYTHONPATH /app
 
-COPY pyproject.toml poetry.lock .
+COPY pyproject.toml poetry.lock README.md .
```
brassy-endomorph, line 15: Is this necessary for the install?

Dockerfile.prod
```diff
@@ -18,7 +18,7 @@ RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
 ENV PATH="/root/.cargo/bin:${PATH}"
 
 # Install Poetry dependencies
-COPY poetry.lock pyproject.toml /app/
+COPY poetry.lock pyproject.toml README.md /app/
```
brassy-endomorph, line 21: Same Q: Is this necessary for the install?

Makefile
```diff
@@ -9,7 +9,6 @@ help: ## Print the help message
 .PHONY: install
 install:
 	poetry install
-	npm install
```
brassy-endomorph, line 12: We can fully remove this `make` target since if we're requiring Docker for the dev env

Makefile
```diff
@@ -29,13 +28,13 @@ lint: ## Lint the code
 	poetry run ruff format --check && \
 	poetry run ruff check && \
 	poetry run mypy .
-	npx prettier --check .
+	docker run --rm -v $(shell pwd):/work tmknom/prettier:3.2.5 --check ./*.md ./docs ./.github/workflows/* ./hushline
```
brassy-endomorph, line 31: This can be done with `docker compose run --rm ` on the dev container if we install NPM there. Which might be better that pulling in some other container? Idk :shrug:

jeremywmoore: I'd rather not include an extra runtime in our application container if it's only used during development.

Dockerfile.dev
jeremywmoore: I saw an error message from poetry complaining that this file was missing, but adding it makes it even more unhappy. I think not including the source for the app before the `poetry install` in the Dockerfile is making poetry 😒

brassy-endomorph: Can you paste the error? I wasn't having any issues with this

Makefile
brassy-endomorph: We have split dev/prod containers, so adding that to the dev container shouldn't conflict with your wishes. Not that we have to add it, but was just sayin'

Makefile
jeremywmoore: I ended up pulling ruff/mypy out of this PR due to an error I encountered when running on CI

Makefile
```diff
@@ -29,13 +28,13 @@ lint: ## Lint the code
 	poetry run ruff format --check && \
 	poetry run ruff check && \
 	poetry run mypy .
-	npx prettier --check .
+	docker compose run --rm -v $(shell pwd):/work app npx prettier --check ./*.md ./docs ./.github/workflows/* ./hushline
```
brassy-endomorph, line 31: when running with `docker compose` you don't have to re-specify the volumes as they already get pulled in. The `workdir` will be `/app` and the repo root will be mounted at `/app`, so this should do what we need without a second mount, no?

brassy-endomorph (approved):

PR #612, opened 2024-09-20
Cleanup migrations

### LINKED ISSUE #591: Cleanup DB migrations
Followup to #272 

- [ ] Ensure consistent naming conventions (also noted in #586)
- [ ] Ensure all tables have an explicit `__tablename__`
- [ ] Ensure all tables name are plural (for consistency)
- [ ] Remove empty revision `c2b6eff6e213_.py` (supersedes #547)
- [ ] Ensure (if possible) linear migrations (for now) to support easier auto-testing of migrations (relates to #555)
- [ ] Implement [Flask-Migrate](https://flask-migrate.readthedocs.io/en/latest/index.html) to give us a nice CLI for playing with migrations in dev

brassy-endomorph:
fixes #591

- adds `Metadata` to `SqlAlchemy` extension so we have explicitly named columns
- ensures all tables have an explicit and plural name
- creates a linear history of of revisions
- creates a new revision to ensure all new standards are applied to prod


hushline/model.py
```diff
@@ -232,15 +231,6 @@ class AuthenticationLog(Model):
     otp_code: Mapped[Optional[str]] = mapped_column(db.String(6))
     timecode: Mapped[Optional[int]]
 
-    __table_args__ = (
```
jeremywmoore, line 235: why remove this? Without this I think alembic will attempt to remove the index. If the names no longer match, can we update?

migrations/versions/5ffe5a5c8e9a_update_entities_to_match_metadata.py
```diff
@@ -0,0 +1,40 @@
...
+down_revision = "46aedec8fd9b"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.execute(sa.text("ALTER INDEX usernames_username_key RENAME TO uq_usernames_username"))
+
+    with op.batch_alter_table("authentication_logs", schema=None) as batch_op:
+        batch_op.drop_index("idx_authentication_logs_user_id_timestamp_successful")
```
jeremywmoore, line 23: why remove this index?

jeremywmoore (commented):
I feel a bit funny about rewriting all this history. Is this going to cleanly apply without any loss of data on the live DB?

hushline/model.py
brassy-endomorph: That index isn't actually used for queries, so it can be safely removed without impacting performance.

migrations/versions/5ffe5a5c8e9a_update_entities_to_match_metadata.py
brassy-endomorph: See other comment

hushline/model.py
brassy-endomorph: I lied lol. `EXPLAIN` was giving me a full table scan when I didn't have enough representative data in it, but after packing the table it uses an index only scan. I'll back this out.

brassy-endomorph:
> I feel a bit funny about rewriting all this history. Is this going to cleanly apply without any loss of data on the live DB?

Do you mean about making the history linear? No, Alembic won't do naughty things if it can't find the right revision. At worst, we have to `flask db stamp $revision_id`

Or do you mean about updating the index/key names in the historic migrations? I used the previous set of revisions and applied them one at a time to get that. Because we added the `metadata` to `SqlAlchemy`, it changes how indexes without names are generated and it broke historic migrations. Now that they've been pinned to what the old migrations would have done, it's safe to apply the migrations with the new metadata.

Also, one of the things before we roll this out that I was going to do was write annoyingly pedantic tests including with data, so by the time any of the new migrations and updates to existing migrations hit prod, we'll have much stronger assurance that they do the right thing.

jeremywmoore (approved):
In reviewing this PR, I create a fresh DB instance using the migrations in `main`, then switch to this branch and applied migrations again. The database seemed in a good state and the app was happy.

PR #618, opened 2024-09-25
Auto testing of migrations

### LINKED ISSUE #555: Alembic migrations should have associated test cases

Migrations can be very error prone, and unusual or unexpected state on a prod server can cause migrations to fail. We can try to anticipate these and write tests cases to ensure migrations are always successful.

SecureDrop has a strategy in [`test_alembic.py`](https://github.com/freedomofpress/securedrop/blob/dbdccc593e1aeab149dbd6019709394ec1e6dc9c/securedrop/tests/test_alembic.py) and the related [`tests/migrations`](https://github.com/freedomofpress/securedrop/tree/dbdccc593e1aeab149dbd6019709394ec1e6dc9c/securedrop/tests/migrations) directory.

**Describe alternatives you've considered**

I do not know of any other ways people test Alembic migrations. More research might lead us to a better solution.

**Additional context**

I have seen so many "lgtm" migrations fail in prod :(( the problem is real.

brassy-endomorph:
Automatically test all (new) migrations.

This is a WIP for initial review. I'd forgotten what a pain in the ass it is to write these kinds of tests, so I wanted to get a sensibility check before I finished.

fixes #555

### DIFF
--- a/hushline/__init__.py
+++ b/hushline/__init__.py
@@ -6,14 +6,13 @@ from typing import Any
 
 from flask import Flask, flash, redirect, request, session, url_for
 from flask.cli import AppGroup
-from flask_migrate import Migrate
 from jinja2 import StrictUndefined
 from sqlalchemy.exc import ProgrammingError
 from werkzeug.middleware.proxy_fix import ProxyFix
 from werkzeug.wrappers.response import Response
 
 from . import admin, premium, routes, settings
-from .db import db
+from .db import db, migrate
 from .model import HostOrganization, Tier, User
 from .version import __version__
 
@@ -79,9 +78,8 @@ def create_app() -> Flask:
         app.logger.info("Development environment detected, enabling jinja2.StrictUndefined")
         app.jinja_env.undefined = StrictUndefined
 
-    # Run migrations
     db.init_app(app)
-    Migrate(app, db)
+    migrate.init_app(app, db)
 
     # Initialize Stripe
     if app.config["STRIPE_SECRET_KEY"]:
--- a/hushline/db.py
+++ b/hushline/db.py
@@ -1,3 +1,4 @@
+from flask_migrate import Migrate
 from flask_sqlalchemy import SQLAlchemy
 from sqlalchemy import MetaData
 
@@ -12,3 +13,4 @@ metadata = MetaData(
 )
 
 db = SQLAlchemy(metadata=metadata)
+migrate = Migrate()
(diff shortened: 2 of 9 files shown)


jeremywmoore:
I love the ideas presented in this PR, but I wonder if this is trying to do too much, mainly around the UpgradeTester class. Looking at this class I don't have a good intuition around how this will evolve over time as schemas evolve.  Maybe we could start out by establishing the goals of this PR? For example, ensuring linear migration history seems great. Another might be that upgrades and downgrades of each revision are applied successfully and that the schema after upgrade + downgrade is unchanged.  Maybe a test that checks that the schema generated by migrations matches the dev schema?

If we did want to go to the point of checking data consistency after each upgrade, a dedicated/static test generated for each migration might be easier to maintain in the long run because it would only need to cover a specific migration and not worry about a changing data model.  But even that feels like it could be an outside amount of work for the value.

brassy-endomorph:
The `Up/DowngradeTester` classes don't evolve. One is written per revision, and they're loaded as tests dynamically. Each tester has a `load_data` function that loads a realistic set of data for that revision and then has the up/downgrade applied. The class holds data so it can run a check to ensure that the specific parts of the migration we expect to break can be tested. This is also the part of the migration that tends to break the most often, not the schemas matching. I can write tests for that too, but this was taking enough time as it is, and this type of test will be very necessary if we want to switch to UUIDs since that has a much biggest chance of everything going wrong.

jeremywmoore:
> The `Up/DowngradeTester` classes don't evolve. One is written per revision, and they're loaded as tests dynamically. Each tester has a `load_data` function that loads a realistic set of data for that revision and then has the up/downgrade applied. The class holds data so it can run a check to ensure that the specific parts of the migration we expect to break can be tested. This is also the part of the migration that tends to break the most often, not the schemas matching. I can write tests for that too, but this was taking enough time as it is, and this type of test will be very necessary if we want to switch to UUIDs since that has a much biggest chance of everything going wrong.

Thanks for the clarification, that makes more sense than my initial read. I think this is a good approach. In general, I assume we'd only expect to write migration test when we're updating existing schemas but not necessarily for additive migrations?

brassy-endomorph:
Yes that's reasonable. We can add an array of "ignorable" migrations that get filtered out once that becomes an issue.

micahflee (approved):
Nice! I just read through the code, and this all makes sense. It definitely feels like a lot more work to write migrations because we also have to test them, but also they should be much more robust after this.

PR #622, opened 2024-09-28
Customer branding 533 v2

### LINKED ISSUE #533: Customer Branding

brassy-endomorph:
App branding

- allows admins of personal servers to update brand settings (name and color)
- disallows anyone of multi-tenant serves from updating brand settings
- adds a migration that loads an initial row to the DB with current defaults

TODOs / follow-ups

- need to fix tabs in `Settings` page because they overflow off the screen and are unclickable

Supersedes #573
Fixes #533


hushline/settings/__init__.py
```diff
@@ -605,6 +616,38 @@ def update_smtp_settings() -> Response | str:
...
+            return redirect(url_for(".index"))
+
+        flash("⛔ Invalid form data. Please try again.")
+        return redirect(url_for(".index"))
+
+    @bp.route("/update-brand-app-name", methods=["POST"])
+    @admin_authentication_required
+    @personal_server_required
+    def update_brand_app_name() -> Response | str:
+        host_org = HostOrganization.fetch_or_default()
```
rmlibre, line 639: I think this revision may swallow DB inconsistency errors during runtime. Perhaps the original with your added ``fetch`` instead of the previous ``default``?

https://github.com/scidsg/hushline/blob/06042b73b1b8367c9a117296d3addc114357e772/hushline/settings.py#L742-L744

hushline/settings/__init__.py
```diff
@@ -605,6 +616,38 @@ def update_smtp_settings() -> Response | str:
         flash("👍 SMTP settings updated successfully")
         return redirect(url_for(".index"))
 
+    @bp.route("/update-brand-primary-color", methods=["POST"])
+    @admin_authentication_required
+    @personal_server_required
+    def update_brand_primary_color() -> Response | str:
+        host_org = HostOrganization.fetch_or_default()
```
rmlibre, line 623: I think this revision may swallow DB inconsistency errors during runtime. Perhaps the original with your added ``fetch`` instead of the previous ``default``?

https://github.com/scidsg/hushline/blob/06042b73b1b8367c9a117296d3addc114357e772/hushline/settings.py#L725-L727

rmlibre (changes requested):
This looks really good, thank you for these updates. I left a few notes

hushline/settings/__init__.py
brassy-endomorph: I don't think we need to error here because if the row isn't present, then the app would be busted in many other places and the user wouldn't even be able to navigate to this page in the first place. Like we don't `try/except` on the DB query itself for `OperationalError`s either, just assume that's live and let the top level error handler deal with that.

hushline/settings/__init__.py
brassy-endomorph: see responses in thread https://github.com/scidsg/hushline/pull/622#discussion_r1780066196

tests/test_settings.py
```diff
@@ -478,3 +479,77 @@ def test_alias_change_directory_visibility(
...
+
+    @pytest.mark.usefixtures("_authenticated_admin")
+    def test_update_brand_primary_color(self, client: FlaskClient, admin: User) -> None:
+        color = "#acab00"
+        resp = client.post(
+            url_for("settings.update_brand_primary_color"),
+            data={"brand_primary_hex_color": color},
+            follow_redirects=True,
+        )
+        assert resp.status_code == 403
```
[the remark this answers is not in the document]
brassy-endomorph: We currently have unstyled 403 pages, so we'd have to update our error handlers and overall strategy for this to make sense. I think that's outside the scope of this PR.

tests/test_settings.py
```diff
@@ -478,3 +479,77 @@ def test_alias_change_directory_visibility(
...
+
+    @pytest.mark.usefixtures("_authenticated_admin")
+    def test_update_brand_app_name(self, client: FlaskClient, admin: User) -> None:
+        name = "h4cK3rZ"
+        resp = client.post(
+            url_for("settings.update_brand_app_name"),
+            data={"brand_app_name": name},
+            follow_redirects=True,
+        )
+        assert resp.status_code == 403
```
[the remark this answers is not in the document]
brassy-endomorph: See https://github.com/scidsg/hushline/pull/622#discussion_r1780062494

glenn-sorrentino (approved):

glenn-sorrentino (approved):

PR #625, opened 2024-09-29
fix: use app logic to avoid broken migration & apply DRY #622

rmlibre:
When running ``make migrate-dev`` there's an error with the new 'host_organization' table migration. Filling & committing the default record on app startup if it's missing:

- removes the error 
- reduces repetition of magic constants which were given variable names in the app 
- keeps application & migration logic separate 
- and, avoids an incomplete DB state which breaks rendering of CSS from jinja db inserts.

Partial reversion of (https://github.com/scidsg/hushline/pull/622/commits/9cad57194ee0ac72a8a56a53d950bb2da2b64d1e, https://github.com/scidsg/hushline/pull/622/commits/9674b1e06e1dab58fec0f44fa2bc5a04d59c2de3)

Suggested in review https://github.com/scidsg/hushline/pull/622#discussion_r1780064725

For Roadmap Item #533

---

### Extended Error Message Fixed by This PR:

```bash
    poetry run ./scripts/dev_migrations.py
    2024-09-29 17:51:17,772:WARNING:Could not check for existence of HostOrganization
    Traceback (most recent call last):
      File "~/.cache/pypoetry/virtualenvs/hushline-syl4zbji-py3.12/lib/python3.12/site-packages/sqlalchemy/engine/base.py", line 1967, in _exec_single_context
        self.dialect.do_execute(
      File "~/.cache/pypoetry/virtualenvs/hushline-syl4zbji-py3.12/lib/python3.12/site-packages/sqlalchemy/engine/default.py", line 941, in do_execute
        cursor.execute(statement, parameters)
      File "~/.cache/pypoetry/virtualenvs/hushline-syl4zbji-py3.12/lib/python3.12/site-packages/psycopg/cursor.py", line 97, in execute
        raise ex.with_traceback(None)
    psycopg.errors.UndefinedTable: relation "host_organization" does not exist
    LINE 2: FROM host_organization
                 ^

    The above exception was the direct cause of the following exception:

    Traceback (most recent call last):
      File "~/hushline/hushline/__init__.py", line 116, in create_app
        host_org = HostOrganization.fetch()
                   ^^^^^^^^^^^^^^^^^^^^^^^^
      File "~/hushline/hushline/model.py", line 46, in fetch
        return db.session.get(cls, cls._DEFAULT_ID)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "~/.cache/pypoetry/virtualenvs/hushline-syl4zbji-py3.12/lib/python3.12/site-packages/sqlalchemy/orm/scoping.py", line 1060, in get
        return self._proxied.get(
               ^^^^^^^^^^^^^^^^^^
      File "~/.cache/pypoetry/virtualenvs/hushline-syl4zbji-py3.12/lib/python3.12/site-packages/sqlalchemy/orm/session.py", line 3693, in get
        return self._get_impl(
               ^^^^^^^^^^^^^^^
      File "~/.cache/pypoetry/virtualenvs/hushline-syl4zbji-py3.12/lib/python3.12/site-packages/sqlalchemy/orm/session.py", line 3873, in _get_impl
        return db_load_fn(
               ^^^^^^^^^^^
      File "~/.cache/pypoetry/virtualenvs/hushline-syl4zbji-py3.12/lib/python3.12/site-packages/sqlalchemy/orm/loading.py", line 694, in load_on_pk_identity
        session.execute(
      File "~/.cache/pypoetry/virtualenvs/hushline-syl4zbji-py3.12/lib/python3.12/site-packages/sqlalchemy/orm/session.py", line 2362, in execute
        return self._execute_internal(
               ^^^^^^^^^^^^^^^^^^^^^^^
      File "~/.cache/pypoetry/virtualenvs/hushline-syl4zbji-py3.12/lib/python3.12/site-packages/sqlalchemy/orm/session.py", line 2247, in _execute_internal
        result: Result[Any] = compile_state_cls.orm_execute_statement(
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "~/.cache/pypoetry/virtualenvs/hushline-syl4zbji-py3.12/lib/python3.12/site-packages/sqlalchemy/orm/context.py", line 305, in orm_execute_statement
        result = conn.execute(
                 ^^^^^^^^^^^^^
      File "~/.cache/pypoetry/virtualenvs/hushline-syl4zbji-py3.12/lib/python3.12/site-packages/sqlalchemy/engine/base.py", line 1418, in execute
        return meth(
               ^^^^^
      File "~/.cache/pypoetry/virtualenvs/hushline-syl4zbji-py3.12/lib/python3.12/site-packages/sqlalchemy/sql/elements.py", line 515, in _execute_on_connection
        return connection._execute_clauseelement(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "~/.cache/pypoetry/virtualenvs/hushline-syl4zbji-py3.12/lib/python3.12/site-packages/sqlalchemy/engine/base.py", line 1640, in _execute_clauseelement
        ret = self._execute_context(
              ^^^^^^^^^^^^^^^^^^^^^^
      File "~/.cache/pypoetry/virtualenvs/hushline-syl4zbji-py3.12/lib/python3.12/site-packages/sqlalchemy/engine/base.py", line 1846, in _execute_context
        return self._exec_single_context(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "~/.cache/pypoetry/virtualenvs/hushline-syl4zbji-py3.12/lib/python3.12/site-packages/sqlalchemy/engine/base.py", line 1986, in _exec_single_context
        self._handle_dbapi_exception(
      File "~/.cache/pypoetry/virtualenvs/hushline-syl4zbji-py3.12/lib/python3.12/site-packages/sqlalchemy/engine/base.py", line 2355, in _handle_dbapi_exception
        raise sqlalchemy_exception.with_traceback(exc_info[2]) from e
      File "~/.cache/pypoetry/virtualenvs/hushline-syl4zbji-py3.12/lib/python3.12/site-packages/sqlalchemy/engine/base.py", line 1967, in _exec_single_context
        self.dialect.do_execute(
      File "~/.cache/pypoetry/virtualenvs/hushline-syl4zbji-py3.12/lib/python3.12/site-packages/sqlalchemy/engine/default.py", line 941, in do_execute
        cursor.execute(statement, parameters)
      File "~/.cache/pypoetry/virtualenvs/hushline-syl4zbji-py3.12/lib/python3.12/site-packages/psycopg/cursor.py", line 97, in execute
        raise ex.with_traceback(None)
    sqlalchemy.exc.ProgrammingError: (psycopg.errors.UndefinedTable) relation "host_organization" does not exist
    LINE 2: FROM host_organization
                 ^
    [SQL: SELECT host_organization.id AS host_organization_id, host_organization.brand_app_name AS host_organization_brand_app_name, host_organization.brand_primary_hex_color AS host_organization_brand_primary_hex_color
    FROM host_organization
    WHERE host_organization.id = %(pk_1)s::INTEGER]
    [parameters: {'pk_1': 1}]
    (Background on this error at: https://sqlalche.me/e/20/f405)
```

---

### Actions Passing:

- [x] https://github.com/rmlibre/hushline/actions/runs/11093540275/job/30819945877

### DIFF
--- a/hushline/__init__.py
+++ b/hushline/__init__.py
@@ -6,7 +6,6 @@ from typing import Any
 from flask import Flask, flash, redirect, request, session, url_for
 from flask_migrate import Migrate
 from jinja2 import StrictUndefined
-from sqlalchemy.exc import ProgrammingError
 from werkzeug.middleware.proxy_fix import ProxyFix
 from werkzeug.wrappers.response import Response
 
@@ -93,7 +92,10 @@ def create_app() -> Flask:
 
     @app.context_processor
     def inject_host() -> dict[str, HostOrganization]:
-        return dict(host_org=HostOrganization.fetch_or_default())
+        if (host_org := HostOrganization.fetch()) is None:
+            db.session.add(host_org := HostOrganization())
+            db.session.commit()
+        return dict(host_org=host_org)
 
     @app.context_processor
     def inject_is_personal_server() -> dict[str, Any]:
@@ -108,18 +110,5 @@ def create_app() -> Flask:
                 f"http://{app.config['ONION_HOSTNAME']}{request.path}"
             )
             return response
-
-    # we can't
-    if app.config.get("FLASK_ENV", None) != "development":
-        with app.app_context():
-            try:
-                host_org = HostOrganization.fetch()
-            except ProgrammingError:
-                app.logger.warning(
-                    "Could not check for existence of HostOrganization", exc_info=True
-                )
-            else:
-                if host_org is None:
-                    app.logger.warning("HostOrganization data not found in database.")
 
     return app
--- a/migrations/versions/5410668e15ad_add_host_organization_table.py
+++ b/migrations/versions/5410668e15ad_add_host_organization_table.py
@@ -26,11 +26,6 @@ def upgrade() -> None:
         sa.PrimaryKeyConstraint("id", name=op.f("pk_host_organization")),
     )
 
-    op.execute(
-        "INSERT INTO host_organization (id, brand_app_name, brand_primary_hex_color) "
-        "VALUES (1, '🤫 Hush Line', '#7d25c1')"
-    )
-
 
 def downgrade() -> None:
     op.drop_table("host_organization")


brassy-endomorph (changes requested):
I extremely disagree with the logic of using the `context_processor` to manipulate DB state. Stateful changes that aren't handled in endpoints should be handled by migrations, queue workers, or (for dev only) in `dev_data.py`. We could scatter the loading of DB objects throughout the app in a ton of places for various reasons, but that will only make it more complicated to reason about what the app's state is upon startup and when it might change.

When running migrations, there will always be a hiccup where two versions of the app are running against one know DB schema. We don't have zero-downtime DB migrations (but with a small DB it's sub second), so there will always be some `ProgrammingError`s when the app expects a different schema.

The constants in the migration aren't magic numbers per se. They're the constant values snapshotted at the time the migration was created. They are eternal. It will always be be case that migration `5410658e15ad` will have used the values `(1, '🤫 Hush Line', '#7d25c1')` to populate the table. Once that migration is used, it should never be modified. It's just a historic artifact, and the migration module shouldn't reference the app code because the app code will drift. We can't use constants defined in the `HostOrganization` class.

Also the use of `fetch_or_default` prevents broken CSS because a valid `HostOrganization` is always returned for use by the app within the endpoint including within `render_template`. The CSS doesn't depend on the row being in the DB but rather an object being passed to the template.

brassy-endomorph:
> I think your expertise & comfort with database management is leading to a bias preventing you from seeing how you can make this code better, and improve the developer experience.

I've been working with Flask and SqlAlchemy for almost 10 years and Alembic for nearly as long. That time means both experience and also yes biases. Don't get me wrong, there's plenty of things I've kept doing for too long and had to relearn along the way, and just in seeing some of this codebase saw some tricks that I've already pulled into other code I work on. Which is to say, sometimes I'm stubborn, but not always nor by default.

> My suggestion is three lines of code in one place that uses the `HostOrganization` interface provided by the app in the app initializer to initialize the app, and solves my local development error.

I'm not aiming for minimal LOC. I'm aiming for minimal mental overhead for long term maintainability, and I'm specifically concerned about having too many places where we initialize data or mutate state.

But to the specific points.

> Your suggestion first inserts hard-coded values here requiring the use of docker and migration tools because it's not taking advantage of the `HostOrganization` interface:

This is how Alembic migrations always should work. They should never rely on app code because migrations may be run long after the app code changes. As in, revision `123` is created on 2023-01-01 with the app at commit `abc`, and it references the app code (`Flask` object, `SqlAlchemy` models, `utils.py`, who knows). A year later, the app code has gone through many iterations and is now at commit `def`. An admin decides to finally update their code and runs the newest version which tries to upgrade to the latest `alembic_version` at revision `456`. If `123` depends on app code, we have to always test revision `123` to ensure that none of the constants, utilities, models, or anything else changed along the way. This is significant overhead and error prone. Basically you repeat yourself once (copy/paste constants to the migration)so that you never have to update that code. Migrations can persist unaltered for years. The longest I've seen is 7.

> But, because your suggestion doesn't work to solve my error, the defaults also have to be manually inserted here:

My original suggestion was to have a `settings` table with a primary key `name` pointing at a JSONB of `value`. This would be a SqlAlchemy model of type `Setting`. The `HostOrganization` would be a wrapper that pulls settings as necessary and handles the default logic because of how constructors don't set default values until insert on `SqlAlchemy` models. This would also allow us to do partial queries where uninitialized settings error if they're accessed instead of returning `None` or similar on a `Model`. But that wasn't an issue and I didn't want to push it because that might be premature optimization. Some of the weirdness of the code you see is because we're using the `Model` a bit like a `dataclass` but that's not quite how it's intended to be used. Or at least, such is my understanding of `Model`s.

> And then because the object the app needs to work still hasn't been committed explicitly, we also have to check and make sure it's committed here:

This was to address the concern that the app should "just work" out of the box, so to account for the possibility that the database row doesn't exist we have to add some extra logic other places.

The advantage of loading data with the schema migration is that we are guaranteed to be in one of two states.

1. The migration applied, the schema updated, and the row is present.
2. The migration failed, the schema is unchanged, and ergo no row could be present at all

Since we don't `try/except` on `ProgrammingError`s for any usage of `HostOrganization` (which is almost every single request), we will never be able to handle a case where the migration didn't apply. If we keep the data migration in with the schema migration, then we can drop the `fetch_or_default` logic and always `fetch` _and_ always know that `SqlAlchemy` is tracking the object and never have to do `session.add(...)`.

Basically at this point of maturity of the app, it is acceptable to not program defensively around database migrations and just accept that a failed migration is a catastrophic error and that if the database doesn't migrate, an unusable app is an acceptable result. A bug, yes, but one that we fix outside of the code.

> Which means if it didn't actually exist, because it wasn't committed explicitly, the admin will have to navigate to the settings -> branding tab on a page without color or an h1 tag to fix the inconsistency.

I had a bug in my PR (I dropped part of a diff while rebasing) where your statement is correct, but I have since fixed that. The way it's programmed in my PR currently (as in, after where you branched), this will never have to happen. The app can completely function if the data migration fails (which it can't without the schema migration also failing). The "inconsistency" would only refer to the lack of a row in the database, but since the app is programmed around that, it isn't an issue.

> But this still isn't enough logic to make sure the object exists in the database, which could just be instantiated and committed during initialization, so your suggestion also checks again here:

This was done to provide exactly one log message on start up if the row isn't there, and it catches the possibility the migration hasn't finished yet so that at least there's a strong hint at what the issue is. This logic being on every request would be undesirable as it would flood the logs on top of what would likely be 500 errors from elsewhere in the app where we don't program defensively around failed migrations.

> I don't know enough about database management to know why, in that context, the database shouldn't be initialized by the app with the records it needs.

This is separation of responsibilities. The app is supposed to start up on its own and quickly start serving requests. This is so that healthchecks can start pinging the app and kill it if it's in a bad state. A migration (schema change, data insert, whatever) being run pre-start up will slow the app and if a migration takes to long, it can exceed the unhealthy time window and the app will be killed before the migration completes. This is what could happen now running `flask db upgrade` in the `start.sh` in the Docker container before `gunicorn "${args[@]}"`. Migrations should happen in a separate container that runs once (as in `replicas: 1`) until it exits 0 that way migrating and the app initializing can happen in parallel. This is also to handle the case where two replicas of the app container start at the same time. `A` sees the DB hasn't been migrated and starts a migration. `B` sees the same and starts a migration. Both are putting load on the DB. `A` finishes success, then `B` errors because on transaction commit, some table/column/whatever has already been updated, thus `B`'s DDL statement fails.

This (app + migration containers) isn't the architecture we have now, but it's typical of most projects I've worked on.

---

Basically we can remove a lot of my code if we don't program defensively around a missing `HostOrganization` and leave it in the Alembic migrations. The `dev_data` container should also load the `HostOrganization` as part of `dev_data.py` and this script should always run on `docker compose up` (currently it doesn't). That would actually be the least LOC to deal with this issue and keeps the app cleaner because it keeps data being loaded / written in only two places: Alembic for prod, `dev_data.py` for dev. That big warning in the `__init__.py` is the ugliest part, and we can solve that too for tests by creating a fixture that autoloads it there.

rmlibre:
We need to do better prioritization. Is it worth the time to engineer the ideal you're envisioning? Or are there more important issues to put our energy towards, like security and privacy problems? Because your suggestion breaks my local dev environment. I provided a minimal patch, which I tested, and gets my environment to work, and I'm hoping to move on to solving better problems. I'm just a little prickly because I just spent the last week debugging, resolving machine breaking application upgrades, and buying new hardware to get my local dev environment to be compatible with the postgres in dev updates you pushed through before I could test them or get #573 merged. And now we're here in contention again about the same thing which you had already passed on blocking in #573.


> A migration (schema change, data insert, whatever) being run pre-start up will slow the app and if a migration takes to long

I don't think, in this case, adding a less than 100 byte record is going to cause the app to timeout. Now that you explain that concern, I understand where you're coming from, but in this case it's not an issue. Planning around the idea that it might be an issue here is wasting our time and breaking my environment.

brassy-endomorph:
> We need to do better prioritization. Is it worth the time to engineer the ideal you're envisioning?

I spent more time writing the previous comment than adding the changes the we're discussing now, and that took longer than my original suggestion of "put it in Alembic" which would probably have been around a 90 second fix.

But also yes I really do think this is worth the discussion because the biggest lesson I've learned is that maintainability is critical especially for FOSS. And that includes mental models of an app like how things are wired up.

> Because your suggestion breaks my local dev environment. I provided a minimal patch, which I tested, and gets my environment to work,

And we should fix that! And there are many ways to do it, and I've also proposed other solutions like ensuring that `dev_data.py` is run by the `dev_data` service in `docker-compose.yml` so that things like manual inserts for the dev env have a dedicated home instead of being put wherever is convenient at the time. Doing that would give us one command to fully spin up the dev env: `docker compose up`. Boom. Done.

> to get my local dev environment to be compatible with the postgres in dev updates that were pushed through before I could test them

I mean, it extremely sucks that this broke your dev env, and I am sorry that happened. I haven't heard of anyone having such and issue before, so I couldn't have anticipated it.

But we were already having drift between dev/test and prod because we didn't enforce foreign keys in Sqlite which means there was a whole class of issues that we could only detect on deployment. Which isn't a a state we want the project in either.

> And now we're here in contention again about the same thing which you had already passed on blocking

Yeah well I thought about it some more and am less convinced that passing on blocking was the right call. I'm still not blocking, however. I'm hoping someone else weighs in here, and if consensus is that we should do it your way, then that's what we'll do until it (if it ever does) becomes a problem.

> I don't think, in this case, adding a less than 100 byte record is going to cause the app to timeout. 

It won't, but that's why migrations run in a separate process that doesn't block app start up. This DML is short and doesn't matter. Longer DML/DDL statements might take seconds or minutes. The point of moving the code into the migrations was not "inserting one row of a few dozen bytes is too slow" but that "all inserts should always be in the migrations."

glenn-sorrentino:
Thanks for the discussion, @rmlibre and @brassy-endomorph. To move this along and improve database management, let's follow Brassy's recommendation. We can always make modifications in the future, and getting branding shipped is a top priority right now.

PR #635, opened 2024-10-03
feat: ensure new passwords are distinct from the previous v2 #468

rmlibre:
# Summary

A simplified and up-to-date v2 for #468, the PR to ensure distinct user passwords during the change password flow.

Adversarial probing of server responses to reveal user password information is prevented. This is done by running basic form validators first, then user password validation, and only then is the repeat password check done.

---

## Passing Actions:

- [x] https://github.com/rmlibre/hushline/actions/runs/11155251787/job/31005889746
- [x] https://github.com/rmlibre/hushline/actions/runs/11155445464/job/31006381720
- [x] https://github.com/rmlibre/hushline/actions/runs/11163047766/job/31029062714
- [ ] https://github.com/rmlibre/hushline/actions/runs/11188325503/job/31106913233 ⚠️ (http resp code issue)
- [x] https://github.com/rmlibre/hushline/actions/runs/11188325503/job/31108203190


tests/test_settings.py
```diff
@@ -60,28 +60,42 @@ def test_change_username(client: FlaskClient, user: User) -> None:
...
-    )
-    assert response.status_code == 200
-    assert "Password successfully changed. Please log in again." in response.text
-    assert "/login" in response.request.url
-    assert len(new_password_hash := user.password_hash) > 32
-    assert new_password_hash.startswith("$scrypt$")
-    assert original_password_hash not in new_password_hash
-    assert user_password not in new_password_hash
-    assert new_password not in new_password_hash
+    for new_password in [user_password, "", "aB!!", "aB3!", (33 * "aB3!")[:129], 5 * "aB3!"]:
```
brassy-endomorph, line 67: we should use pytest's `parametrize` mark to inject these if possible. it makes it easier to see which of the cases fail

rmlibre: There's two issues with going the parametrize route:

1. Because we still haven't solved the slow test problem, using parametrize can add about a second or more of execution time to the tests for each piece of example data. That's a bit painful.
2. The test could use a refactor, because it's structured to test everything in one big flow. I'd like to have several tests with declarative names, like ``test_empty_new_password_entry_is_invalid``, etc. But because of the slow tests, this better pattern slows down the tests dramatically. This should probably wait until we can speed up the tests.

It's a little verbose, but for now I can add context data to the end of each assert, so it's shown if that line throws.

rmlibre: Addressed by (https://github.com/scidsg/hushline/pull/635/commits/7761f1d6a0b68421bfd918928615b639e6373746)

brassy-endomorph (approved):

brassy-endomorph (approved):

PR #651, opened 2024-10-05
Style updates

glenn-sorrentino:
- Smaller encrypted text
- Fix dark mode destructive colors

Screenshot 2024-10-05 at 12 29 34 PM

**Before:**
Screenshot 2024-10-05 at 12 29 40 PM


scidsg-dev (approved):

scidsg-dev (approved):

PR #676, opened 2024-10-09
Fix Dark Mode Upgrade Banner

glenn-sorrentino:
Before:
Screenshot 2024-10-08 at 7 15 18 PM

After:
Screenshot 2024-10-08 at 7 15 07 PM


scidsg-dev (approved):

PR #678, opened 2024-10-09
Replace is_personal_server flag

### LINKED ISSUE #626: Incorrect and unconfigurable `X-Forwarded-*` proxy rules

We have a repository for the app that is not guaranteed to be in sync with the repository for _our_ production deployment (much less anyone else's). We configure the `ProxyFix` object like so:

https://github.com/scidsg/hushline/blob/5b6322b798407f808686ba8dcfb44e5d127313d2/hushline/__init__.py#L62-L66

This _may_ be correct now, but on first look I'm not convinced it is.

These values should be configurable. If any are set, then we set all values via env vars + defaults and apply the `ProxyFix`. This shouldn't be hardcoded and then only depend on the `IS_PERSONAL_SERVER` env var.

**Additional Context**

I'm not sure we actually depend on any of these values anywhere. As in as far as I can tell, the addresses pulled from the `X-Fowarded-*` are never used in any way. It might be simpler to just yank this entirely and then re-add it when necessary.

### LINKED ISSUE #646: Remove `IS_PERSONAL_SERVER` config in favor of specific flags per feature

The notion of a "personal server" is ambiguous and internally on the Hushline team is used to differentiate between the flagship instance at `tips.hushline.app` and one someone else hosts. The flag `IS_PERSONAL_SERVER` is set to change the behavior of the app.

Now that we have Stripe integration, there are open tickets and discussions that require features to be conditionally gated on payment, and there are additionally open discussions that blend both paid features with "personal server." This is confusing and adds complexity to the logic of the app.

1. Any self-hosted instance should have all features enabled for all users by default without gating on a paid tier
2. Completely remove the notion of "personal server" from the app as there should be no distinction of any sort between the flagship ("ours") and a hosted instance ("theirs"). 
3. For any feature that can be conditionally enabled, have a single feature flag to enable or disable that. These may be boolean or if they require payment could be ternary with values being `enabled`/`disabled`/`paid-only`.

**Additional Context**

Generally speaking, one flag should enable or disable a single feature so that its meaning is totally clear and transparent. Overuse of multi-purpose flags like this can make it hard to reason about what flipping a single flag does. Further, it hides dependencies between different features by lumping them together. If we want to decouple them later, it's far more work rather than explicitly surfacing the complexity at the time a feature is implemented.

That said, it simplifies testing as we only have to test two types of servers (hosted and personal) instead of matrix testing many combinations of configurations.

brassy-endomorph:
fixes #646
fixes #626 

- Removes `IS_PERSONAL_SERVER` flag and `is_personal_server` jinja var
- Removes hardcoded riseup.net text and replaces with `SMTP_FORWARDING_MESSAGE_HTML` env var
- Changes logic for `/directory` route to use `DIRECTORY_VERIFIED_TAB_ENABLED` env var instead of `IS_PERSONAL_SERVER`
- Removes `ProxyFix` that has dubiously correct settings with unused results
- Detects "personal server / Tor" via `request.host`
- renamed JS files (didn't change them, but the diff thinks I did)
- adds docs for all currently supported configs (others exist, but they're purposely undocumented so we don't have to support them)

### DIFF
--- a/hushline/static/js/directory.js
+++ b/hushline/static/js/directory.js
@@ -52,7 +52,7 @@ document.addEventListener("DOMContentLoaded", function () {
   function reportUser(username, bio) {
     const messageContent = `Reported user: ${username}\n\nBio: ${bio}\n\nReason:`;
     const encodedMessage = encodeURIComponent(messageContent);
-    const submissionUrl = `${pathPrefix}/profile/admin?prefill=${encodedMessage}`;
+    const submissionUrl = `${pathPrefix}/to/admin?prefill=${encodedMessage}`;
     window.location.href = submissionUrl;
   }
 
@@ -96,7 +96,7 @@ document.addEventListener("DOMContentLoaded", function () {
                     <div class="badgeContainer">${badgeContainer}</div>
                     ${bioHighlighted ? `<p class="bio">${bioHighlighted}</p>` : ""}
                     <div class="user-actions">
-                        <a href="${pathPrefix}/profile/${user.primary_username}">Send a Message</a>
+                        <a href="${pathPrefix}/to/${user.primary_username}">Send a Message</a>
                         ${isSessionUser ? `<a href="#" class="report-link" data-username="${user.primary_username}" data-display-name="${user.display_name || user.primary_username}" data-bio="${user.bio ?? "No bio"}">Report Account</a>` : ``}
                     </div>
                 `;
--- a/hushline/static/js/directory.js
+++ b/hushline/static/js/directory.js
@@ -1,76 +1,63 @@
 document.addEventListener("DOMContentLoaded", function () {
-  // Get the path prefix
-  // If window.location.pathname is /tips/directory, then prefix is /tips
-  // If it's /directory, then prefix is /
   const pathPrefix = window.location.pathname.split("/").slice(0, -1).join("/");
-  const tabs = document.querySelectorAll(".tab");
-  const tabPanels = document.querySelectorAll(".tab-content");
   const searchInput = document.getElementById("searchInput");
   const clearIcon = document.getElementById("clearIcon");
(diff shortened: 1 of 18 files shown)


hushline/templates/premium.html
glenn-sorrentino, line 1: I removed the `base.html` integration because I needed to assign an extra class to the `<main>` tag. If reverting, can we find a solution to inject a class to the tag?

glenn-sorrentino: Here's a PR for the fix: https://github.com/scidsg/hushline/pull/681/

hushline/static/js/directory.js
```diff
@@ -1,76 +1,63 @@
...
   function reportUser(username, bio) {
-    // Construct the message content with explicit line breaks
     const messageContent = `Reported user: ${username}\n\nBio: ${bio}\n\nReason:`;
-
-    // Encode the message content to ensure line breaks and other special characters are correctly handled
     const encodedMessage = encodeURIComponent(messageContent);
-
-    // Redirect to the message submission form for the admin with the pre-filled content
-    const submissionUrl = `${pathPrefix}/to/admin?prefill=${encodedMessage}`;
+    const submissionUrl = `${pathPrefix}/profile/admin?prefill=${encodedMessage}`;
```
glenn-sorrentino, line 55: the profile URL is `/to/username` not `/profile/username`

glenn-sorrentino (changes requested):
Wow this is amazing work. Left a few comments in the code.

hushline/static/js/directory.js
glenn-sorrentino: Also in #681

glenn-sorrentino (dismissed):

glenn-sorrentino (approved):

PR #681, opened 2024-10-09
Add block to `<main>` in base template for class injection

glenn-sorrentino:
Adds the ability to inject a class into the `<main>` tag. This is to accommodate the tier-selection screen which requires a wider container width.


brassy-endomorph:
I integrated this work into mine because this changeset doesn't fully fix the issues introduced by my changes. See my PR for the full fix.

PR #686, opened 2024-10-10
move config parsing to own module

### LINKED ISSUE #637: Parse env vars into `Config` object before configuring `Flask` object

We currently have a bit of logic in the `create_app` function that handles config parsing and validation, and we should pull this into a separate config object so that we can more easily plug in arbitrary values at runtime without having to rebuild the app.

```python
CONFIG_EXTRA_PREFIX = "X_"
CONFIG_JSON_EXTRA_PREFIX = "X_JSON"

def load_config(env: Mapping[str, str] | None = None) -> Mapping[str, Any]:
  if not env:
    env = dict(os.environ)
  config = {}

  # all our known configs go here

  if "FOO" in env:
    config["RENAMED"] = datetime.strptime("%Y-%M-%d", env["FOO"])
    config["BAR"] = env['BAR']  # here bar depends on foo, but isn't always there

  ...

  # unknown configs that are supposed to be strings
  for k, v in env.items():
    if k.startswith(CONFIG_EXTRA_PREFIX):
      config[k[len(CONFIG_EXTRA_PREFIX):]] = v

  # unknown configs that might need to be boolean or numbers
  for k, v in env.items():
    if k.startswith(CONFIG_JSON_EXTRA_PREFIX):
      try:
        value = json.loads(v)
      except JSONDecodeError:
        raise ValueError("Key {k} has invalid JSON")  # hide original error to not log secrets
      
      config[k[len(CONFIG_JSON_EXTRA_PREFIX):]] = value

  return config

app.config.from_mapping(load_config())
```

This method can easily be extended to look for a) env vars and then b) docker secrets with some clear override pattern. If nothing else, this should at least be pulled into its own module however we do it because there is about to be an explosion of env vars with some upcoming PRs and our config is a bit haphazard currently.

This method (and similar) also have the advantage that we're always working on a `Mapping` and not the raw `os.environ` which means slightly less weird patching and fixtures in the tests.

**Describe alternatives you've considered**

There are so many ways to do this that enumerating them all would be a total pain.

brassy-endomorph:
fixes #637

This is a minimal change to organize the config parsing (somewhat) according to some logic that seemed reasonable to me (for now). There are some slight change in the parsing logic so we don't set empty/`None` values in a bunch of places since this make some logic a little cleaner so we can do tests like `if "KEY" in app.config` which is different than `if app.config.get("KEY")`.

- moves all config parsing to `config.py`
- parses random env vars prefixed with `HL_CFG_`
- parses randm env vars prefixed with `HL_CFG_JSON_` as JSON
- basic tests for parsing
- updated docs
- bumped dependencies because somehow that fixed linting (???)


glenn-sorrentino:
Loggin in causes this:

```
    File "/root/.cache/pypoetry/virtualenvs/hushline-9TtSrW0h-py3.12/lib/python3.12/site-packages/flask/app.py", line 1498, in __call__

    return self.wsgi_app(environ, start_response)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

    File "/root/.cache/pypoetry/virtualenvs/hushline-9TtSrW0h-py3.12/lib/python3.12/site-packages/flask/app.py", line 1476, in wsgi_app

    response = self.handle_exception(e)
               ^^^^^^^^^^^^^^^^^^^^^^^^

    File "/root/.cache/pypoetry/virtualenvs/hushline-9TtSrW0h-py3.12/lib/python3.12/site-packages/flask/app.py", line 1473, in wsgi_app

    response = self.full_dispatch_request()
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^

    File "/root/.cache/pypoetry/virtualenvs/hushline-9TtSrW0h-py3.12/lib/python3.12/site-packages/flask/app.py", line 882, in full_dispatch_request

    rv = self.handle_user_exception(e)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

    File "/root/.cache/pypoetry/virtualenvs/hushline-9TtSrW0h-py3.12/lib/python3.12/site-packages/flask/app.py", line 880, in full_dispatch_request

    rv = self.dispatch_request()
         ^^^^^^^^^^^^^^^^^^^^^^^

    File "/root/.cache/pypoetry/virtualenvs/hushline-9TtSrW0h-py3.12/lib/python3.12/site-packages/flask/app.py", line 865, in dispatch_request

    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)  # type: ignore[no-any-return]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

    File "/app/hushline/routes.py", line 412, in login

    if app.config["STRIPE_SECRET_KEY"]:
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

    KeyError: 'STRIPE_SECRET_KEY'
```

brassy-endomorph:
I didn't actually test it yet. Just got it to pass CI, and this check is whether this method looks sensible. But yes, noted.

glenn-sorrentino (approved):

PR #687, opened 2024-10-10
Removes native bold styling

glenn-sorrentino:
Replaces with `--font-sans-bold`


scidsg-dev (approved):

PR #707, opened 2024-10-25
fixes stripe json response

### LINKED ISSUE #706: Stripe - Cancel Flow
Canceling a subscription redirects to https://tips.hushline.app/premium/disable-autorenew which shows a JSON response rather than redirecting to Settings. 

Any further actions to either auto-renew or cancel immediately redirects to the same URL (disable-autorenew)

Screenshot 2024-10-25 at 12 37 35 PM

glenn-sorrentino:
Fixes #706 

- uses a flash message to return responses. 
- fix template to point to appropriate routes
- refactors tests
- Includes update to intro paragraph to contain a link back to hushline.app.
- Fixes link to point to current docs.

### DIFF
--- a/hushline/templates/directory.html
+++ b/hushline/templates/directory.html
@@ -1,188 +1,172 @@
-{% extends "base.html" %}
-
-{% block title %}Directory{% endblock %}
-
-{% block content %}
-  <h2>User Directory</h2>
-  {% if directory_verified_tab_enabled %}
-    {% if not logged_in %}
-      <p class="helper section-intro">
-        👋 Hush Line connects whistleblowers with lawyers, journalists, business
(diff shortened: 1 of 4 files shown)


micahflee (approved):

hushline/templates/directory.html
```diff
@@ -1,188 +1,172 @@
...
-            aria-selected="false"
-            aria-controls="all"
-            id="all-tab"
-          >
-            All
-          </button>
-        </li>
-      </ul>
-    </div>
+{% extends "base.html" %} {% block title %}Directory{% endblock %} {% block
```
brassy-endomorph, line 1: It would be nice to not auto format the HTML because this is now less readable. Blocks should have nested indents and not be flattened on to one line.

PR #730, opened 2024-10-31
Logo upload

brassy-endomorph:
This is a draft PR for general functionality before polishing

- Adds flask extension for storing data in a public directory/bucket
- Configurable drivers for using FS or S3
- upload/delete a user logo
- Split the settings into a key/value table because otherwise every future addition will be a migration
- DB migrations
- Fixed tests due to migrations
- Wrote basic tests for logo upload
- Docs for new config options

TODOs

- Make it less ugly


hushline/templates/settings/branding.html
glenn-sorrentino, line 1: Can we place the delete button in the logo upload form above it?

brassy-endomorph:
@micahflee I'm tempted to merge this as is and leave that migration change out because:

```
$ rg stripe_customer_id
hushline/premium.py
169:    if user.stripe_customer_id is not None:
171:            return stripe.Customer.modify(user.stripe_customer_id, email=email)
173:            user.stripe_customer_id = None
176:    user.stripe_customer_id = stripe_customer.id
208:        db.select(User).filter_by(stripe_customer_id=subscription.customer)
483:                customer=user.stripe_customer_id,

hushline/model.py
238:    stripe_customer_id = mapped_column(db.String(255))
541:            db.select(User).filter_by(stripe_customer_id=invoice.customer)
```

I looks like we're def filtering on that column in a way an index makes sense, so we should add it back in in another PR.

hushline/templates/settings/branding.html
brassy-endomorph: done

glenn-sorrentino (approved):

micahflee (approved):

PR #738, opened 2024-11-05
Add Header Background Translucency

glenn-sorrentino:
Updates the CSS treatment for the header to add translucency to give the app more depth and sense of space.


scidsg-dev (approved):

scidsg-dev (approved):

PR #748, opened 2024-11-09
Fix `delete_message` bug

### LINKED ISSUE #747: Messages not deleting
"Message not found" when trying to delete a message in my inbox.

glenn-sorrentino:
This PR fixes #747 - a bug in the `delete_message` route where messages could not be deleted due to incorrect conditions in the `where` clause, resulting in “Message not found” errors even for valid requests. The bug fix includes modifying the query to properly associate messages with the authenticated user’s primary `Username`.

## Changes Made

1. Bug Fix in `delete_message` Route:
	• Adjusted the `where` clause in the `delete_message` function to check for the `correct username_id` associated with the authenticated user’s `User` object.
	• Added logging statements to aid in troubleshooting future issues with message deletion, including logging the `message_id` and `user_id`.
2. Unit Test for delete_message Route:
	• Created a new unit test (`test_delete_message`) in `tests/test_inbox.py` to validate the `delete_message` route functionality.
	• The test does the following:
	• Sets up a primary `Username` for the user and associates a `Message` with this username.
	• Confirms that the message exists before deletion.
	• Sends a POST request to the `delete_message` endpoint and verifies that the message is successfully deleted.
	• Checks that the correct success message (“Message deleted successfully”) is displayed in the response.

## Testing

• Ran the new test and confirmed that it passes, verifying that the delete_message route is functioning as expected.
• Confirmed that messages are properly deleted from the database and that relevant success messages are displayed to the user.

### DIFF
--- a/hushline/routes.py
+++ b/hushline/routes.py
@@ -288,25 +288,30 @@ def init_app(app: Flask) -> None:
     @app.route("/delete_message/<int:message_id>", methods=["POST"])
     @authentication_required
     def delete_message(message_id: int) -> Response:
+        # Check if the user is logged in
         if "user_id" not in session:
             flash("🔑 Please log in to continue.")
             return redirect(url_for("login"))
 
+        # Fetch the user based on session user_id
         user = db.session.get(User, session.get("user_id"))
         if not user:
             flash("🫥 User not found. Please log in again.")
             return redirect(url_for("login"))
 
+        # Dynamically fetch all username IDs associated with the user
+        usernames_ids = db.session.scalars(
+            db.select(Username.id).where(Username.user_id == user.id)
+        ).all()
+
+        # Attempt to delete the message for this user
         row_count = db.session.execute(
             db.delete(Message).where(
-                Message.id == message_id,
-                Message.username_id.in_(
-                    db.select(Username.user_id)
-                    .select_from(Username)
-                    .filter(Username.user_id == user.id)
-                ),
+                Message.id == message_id, Message.username_id.in_(usernames_ids)
             )
         ).rowcount
+
+        # Match case to determine action based on row count
         match row_count:
             case 1:
                 db.session.commit()
@@ -321,6 +326,7 @@ def init_app(app: Flask) -> None:
                 )
                 flash("Internal server error. Message not deleted.")
 
+        # Redirect to the inbox after deletion attempt
         return redirect(url_for("inbox"))
 
     @app.route("/register", methods=["GET", "POST"])
--- a/hushline/routes.py
+++ b/hushline/routes.py
@@ -301,9 +301,7 @@ def init_app(app: Flask) -> None:
             db.delete(Message).where(
                 Message.id == message_id,
                 Message.username_id.in_(
-                    db.select(Username.id)
-                    .select_from(Username)
-                    .filter(Username.user_id == user.id)
+                    db.select(Username.id).select_from(Username).filter(Username.user_id == user.id)
                 ),
             )
         ).rowcount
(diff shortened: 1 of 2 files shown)


hushline/routes.py
```diff
@@ -288,25 +288,30 @@ def redirect_submit_message(username: str) -> Response:
...
             flash("🔑 Please log in to continue.")
             return redirect(url_for("login"))
 
+        # Fetch the user based on session user_id
         user = db.session.get(User, session.get("user_id"))
         if not user:
             flash("🫥 User not found. Please log in again.")
             return redirect(url_for("login"))
 
+        # Dynamically fetch all username IDs associated with the user
```
brassy-endomorph, line 302: This can be done in a single query if we undo my typo:

```diff
diff --git a/hushline/routes.py b/hushline/routes.py
index a111245..aae59f7 100644
--- a/hushline/routes.py
+++ b/hushline/routes.py
@@ -301,7 +301,7 @@ def init_app(app: Flask) -> None:
             db.delete(Message).where(
                 Message.id == message_id,
                 Message.username_id.in_(
-                    db.select(Username.user_id)
+                    db.select(Username.id)
                     .select_from(Username)
                     .filter(Username.user_id == user.id)
                 ),
```

tests/test_inbox.py
```diff
@@ -0,0 +1,32 @@
+import pytest
+from flask import url_for
+from flask.testing import FlaskClient
+
+from hushline.db import db
+from hushline.model import Message, User, Username
+
+
+@pytest.mark.usefixtures("_authenticated_user")
+def test_delete_message(client: FlaskClient, user: User) -> None:
```
brassy-endomorph, line 10: if we're adding tests for this it would be good to make sure that User A can't delete User B's messages, but this isn't a blocker to merging.

tests/test_inbox.py
```diff
@@ -0,0 +1,32 @@
...
+    # Setup: create primary and alias usernames, and a message for the user
+    primary_username = Username(
```
brassy-endomorph, line 12: Usernames and messages should be created via fixture since there's not any special logic here.

hushline/routes.py
glenn-sorrentino: ahhhhhhhh.

glenn-sorrentino: reverted, updated typo

tests/test_inbox.py
glenn-sorrentino: done

tests/test_inbox.py
glenn-sorrentino: done

tests/test_inbox.py
```diff
@@ -0,0 +1,94 @@
...
+from hushline.db import db
+from hushline.model import Message, User, Username
+
+# Define test passwords
+TEST_USER_PASSWORD = os.getenv("TEST_USER_PASSWORD", "TestPassword123!")
+OTHER_USER_PASSWORD = os.getenv("OTHER_USER_PASSWORD", "OtherUserPass456!")
+
+
+@pytest.fixture()
+def setup_user_data(user: User) -> None:
```
brassy-endomorph: We already have fixtures `user_alias` and `message` that would do this. Is there something here that isn't covered by those?

tests/test_inbox.py
```diff
@@ -0,0 +1,94 @@
...
+        content="Test message for deletion.",
+    )
+    db.session.add(message)
+    db.session.commit()
+
+    return primary_username, alias_username, message
+
+
+@pytest.fixture()
+def other_user() -> User:
```
brassy-endomorph, line 50: This can be added to `conftest` as `user_2` which would establish a repeatable naming pattern

glenn-sorrentino: done

tests/test_inbox.py
glenn-sorrentino: updated

tests/test_inbox.py
brassy-endomorph: Did you drop a commit? They're still in the test file

glenn-sorrentino: This is outdated.

PR #749, opened 2024-11-10
Add Field for the Directory Intro Paragraph

glenn-sorrentino:
This PR adds a new feature that allows administrators to set a customizable introductory description for the user directory. This introductory text appears at the top of the directory page, providing context or guidelines to visitors and enhancing user engagement.

## Key Changes:
- Introduced a `DIRECTORY_INTRO` setting in `OrganizationSetting` to store the introductory text for the user directory.
- Added a form in the admin settings page that allows administrators to update the `DIRECTORY_INTRO` text.
- Implemented a new route `update_directory_intro_text` in the settings module to handle updates to the introductory text by admins.
- Modified the directory template to display the `DIRECTORY_INTRO` text at the top of the directory page for all users.
- Added unit tests to verify that the introductory text can be updated successfully and is displayed correctly on the directory page.


hushline/routes.py
```diff
@@ -518,10 +526,15 @@ def logout() -> Response:
     @app.route("/directory")
     def directory() -> Response | str:
         logged_in = "user_id" in session
+        # Provide a default intro text if DIRECTORY_INTRO is not set
```
brassy-endomorph, line 529: nit: unnecessary comment. there's a bunch of these actually and they should be yanked because they simply state exactly what the line is doing whereas comments should be to explain business logic or clarify code that cannot be made less convoluted.

hushline/settings/__init__.py
```diff
@@ -235,6 +236,15 @@ async def index() -> str | Response:
             flash("🫥 User not found.")
             return redirect(url_for("login"))
 
+        # Initialize the directory intro text form with existing value
+        directory_intro_text_value = (
+            OrganizationSetting.fetch_one(OrganizationSetting.DIRECTORY_INTRO).value
```
brassy-endomorph, line 241: This is making two DB calls. Assign a temp variable and then do your if/else logic.

hushline/model.py
```diff
@@ -40,6 +40,7 @@ class OrganizationSetting(Model):
     BRAND_LOGO = "brand_logo"
     BRAND_NAME = "brand_name"
     BRAND_PRIMARY_COLOR = "brand_primary_color"
+    DIRECTORY_INTRO = "directory_intro"
```
brassy-endomorph, line 43: It might be worth adding a suffix like `_html` or `_markdown` or `_text` to make it more clear to us when we use wit what it's supposed to be. I assume we want `html` but would need to use something like [`bleach`](https://pypi.org/project/bleach/) to sanitize it to a same minimal subset of HTML.

hushline/templates/directory.html
```diff
@@ -2,14 +2,14 @@
 content %}
 <h2>User Directory</h2>
 {% if directory_verified_tab_enabled %} {% if not logged_in %}
+{% if intro_text %}
+<p class="helper section-intro">{{ intro_text }}</p>
```
brassy-endomorph, line 6: nit: indentation so that if/else block is more clear

hushline/templates/settings/branding.html
```diff
@@ -40,6 +40,18 @@ <h3>App Name</h3>
...
   </form>
 
+  <h3>Directory Introduction Text</h3>
+  <form
+    method="POST"
+    action="{{ url_for('settings.update_directory_intro_text') }}"
+    class="formBody"
+  >
+    {{ update_directory_intro_text_form.hidden_tag() }}
+    <label for="directory-intro-text">Introduction Text</label>
```
brassy-endomorph, line 50: this can be achieved with `{{ update_directory_intro_text_form.directory_intro_text.label }}`

tests/test_settings.py
```diff
@@ -666,3 +666,45 @@ def test_update_brand_logo(client: FlaskClient, admin: User) -> None:
     resp = client.get(logo_url, follow_redirects=True)
     # yes this check is ridiculous. why? because we redirect not-founds instead of actually 404-ing
     assert "That page doesn" in resp.text
+
+
+@pytest.mark.usefixtures("_authenticated_admin")
+def test_update_directory_intro_text(client: FlaskClient, admin: User) -> None:
+    """Test updating the directory intro text and ensure data integrity on failure."""
+
+    # Step 1: Set initial intro text and confirm it loads correctly
```
brassy-endomorph, line 675: It might be better to have this test check the endpoint when the value isn't set to we know that a missing/`null` value in the DB doesn't break things.

hushline/routes.py
```diff
@@ -518,10 +526,15 @@ def logout() -> Response:
...
+        intro_text_setting = OrganizationSetting.fetch_one(OrganizationSetting.DIRECTORY_INTRO)
+        intro_text = intro_text_setting.value if intro_text_setting else "Welcome to the directory."
```
brassy-endomorph, line 531: this if/else will not handle the case where there is empty/null intro text. we likely want:

```python
if  (intro_text_setting := OrganizationSetting.fetch_one(OrganizationSetting.DIRECTORY_INTRO)) \
    and intro_text_setting.value:
    intro_text = intro_text_setting.value
else:
    intro_text = "Welcome to the directory."
```

tests/test_settings.py
```diff
@@ -666,3 +666,45 @@ def test_update_brand_logo(client: FlaskClient, admin: User) -> None:
...
+        follow_redirects=True,
+    )
+    assert response.status_code == 200
+    assert "✅ Directory introduction text updated successfully." in response.text
+
+    # Confirm intro text was successfully updated in the database
+    setting = OrganizationSetting.fetch_one(OrganizationSetting.DIRECTORY_INTRO)
+    assert setting.value == updated_intro_text
+
+    # Step 3: Test invalid update (e.g., empty intro text), confirm flash message,
```
brassy-endomorph, line 698: We might want to allow setting an empty string and then say in the field label that an empty string deletes and resets to the original. Or we add a delete button. Idk.

brassy-endomorph (changes requested):

hushline/routes.py
glenn-sorrentino: ✅

hushline/settings/__init__.py
glenn-sorrentino: In #40eb92c

hushline/templates/settings/branding.html
glenn-sorrentino: 7bd531d

hushline/routes.py
glenn-sorrentino: 11d994c

hushline/model.py
glenn-sorrentino: sounds good but lets hold until we can rationalize applying this pattern holistically.

tests/test_settings.py
glenn-sorrentino: updated with the behavior to remove the p tag with no field content

hushline/settings/__init__.py
```diff
@@ -54,9 +55,19 @@
     UpdateBrandAppNameForm,
     UpdateBrandLogoForm,
     UpdateBrandPrimaryColorForm,
+    UpdateDirectoryIntroTextForm,
 )
 
 
+def sanitize_input(input_text: str) -> str:
```
brassy-endomorph, line 62: We should add this to the docs that these are the allowed fields/attributes. Possibly also in the UI.

hushline/settings/__init__.py
```diff
@@ -800,4 +821,27 @@ async def alias(username_id: int) -> Response | str:
             profile_form=profile_form,
         )
 
+    @bp.route("/update-directory-intro", methods=["POST"])
+    @admin_authentication_required
+    def update_directory_intro_text() -> Response:
+        form = UpdateDirectoryIntroTextForm()
+        if form.validate_on_submit():
+            # Get the raw input from the form
```
brassy-endomorph, line 829: These should be collapsed into just two significant lines since the meaning of each is exceedingly clear. More lines and temp vars is more to keep track of mentally

```python
OrganizationSetting.upsert(
    key=OrganizationSetting.DIRECTORY_INTRO,
    value=Markup(sanitize_input(from.directory_intro_text.data)),
)
db.session.commit()
```

hushline/templates/directory.html
```diff
@@ -169,4 +113,4 @@ <h3>{{ username.display_name or username.username }}</h3>
 <script src="{{ url_for('static', filename='js/directory_verified.js') }}"></script>
 {% else %}
 <script src="{{ url_for('static', filename='js/directory.js') }}"></script>
-{% endif %} {% endblock %}
+{% endif %} {% endblock %}
```
brassy-endomorph, line 116: Nit: Can you configure your IDE to leave (and/or add) trailing newlines since that's part of the POSIX standard

pyproject.toml
```diff
@@ -27,6 +27,8 @@ stripe = "^10.9.0"
 types-requests = "^2.32.0.20240712"
 types-setuptools = "^71.1.0.20240813"
 python = "^3.11"
+bleach = "^6.2.0"
```
brassy-endomorph, line 30: `bleach` should be moved up to be alphabetized. `types-bleach` should be moved into dev dependencies only.

hushline/templates/directory.html
```diff
@@ -1,80 +1,42 @@
...
-{% if directory_verified_tab_enabled %} {% if not logged_in %}
-<p class="helper section-intro">
-  👋
-  <a href="https://hushline.app" target="_blank">Hush Line</a>
-  connects whistleblowers with lawyers, journalists, business leaders, and more.
-  We advise speaking to a legal professional before taking any confidential
-  information from your workplace.
-</p>
+{% if intro_text %}
+  <p class="helper section-intro">{{ intro_text | safe }}</p>
```
brassy-endomorph, line 5: We shouldn't use the `safe` filter. The output of the sanitizer should be immediately wrapped in a `Markup` so that we get (semi-) type safety and can't accidentally fuck up later and have a plain `str` marked as `safe`.

hushline/routes.py
```diff
@@ -516,10 +524,23 @@ def logout() -> Response:
...
+            )
+            and intro_text_setting.value
+            else None
+        )
+
         return render_template(
             "directory.html",
             usernames=get_directory_usernames(),
             logged_in=logged_in,
+            intro_text=intro_text,
```
brassy-endomorph, line 543: Intro text must be escaped here also. User input into the DB should be (almost) always be treated as untrusted. If we want to pre-clean/sanitize it going in, that's fine, but at display time we always need to do it again. Also wrap it in `Markup`

brassy-endomorph (changes requested):
A few nits but mostly a note about preventing XSS. Data needs to be escaped at display time always unless we can guaranteed it is safe. We usually can't. A rule of thumb is "user input is lava" and we want to handle it very very carefully. We don't even mark ints or UUIDs as `safe` when we display them even though we can assume quite strongly that they're always going to be safe. HTML that users input is never going be that, so when we pull from the DB it always needs to be cleaned/sanitized and then marked with a safe type.

pyproject.toml
glenn-sorrentino: ✅

hushline/templates/directory.html
glenn-sorrentino: ✅

hushline/routes.py
glenn-sorrentino: 323efb6

hushline/templates/directory.html
glenn-sorrentino: ✅

hushline/routes.py
```diff
@@ -51,6 +62,13 @@ def get_directory_usernames() -> Sequence[Username]:
     ).all()
 
 
+def directory() -> str:
```
brassy-endomorph, line 65: What does this function do? In the routes it's being overwritten by another `def directory()`? It looks like it's fully unused.

poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 1.8.3 and should not be changed by hand.
+# This file is automatically @generated by Poetry 1.8.4 and should not be changed by hand.
```
brassy-endomorph, line 1: You need to run `poetry lock` because i'm getting an error on your branch when i try to build the container:

```
 => ERROR [app 9/9] RUN poetry install                                                   1.0s
------                                                                                        
 > [app 9/9] RUN poetry install:                                                              
0.558 Creating virtualenv hushline-9TtSrW0h-py3.12 in /root/.cache/pypoetry/virtualenvs       
0.818 Installing dependencies from lock file                                                  
0.917 
0.917 pyproject.toml changed significantly since poetry.lock was last generated. Run `poetry lock [--no-update]` to fix the lock file.
------
failed to solve: process "/bin/sh -c poetry install" did not complete successfully: exit code: 1
```

hushline/settings/__init__.py
```diff
@@ -800,4 +823,25 @@ async def alias(username_id: int) -> Response | str:
...
+    def update_directory_intro_text() -> Response:
+        form = UpdateDirectoryIntroTextForm()
+        if form.validate_on_submit():
+            raw_intro_text = form.directory_intro_text.data
+            sanitized_intro_text = sanitize_input(raw_intro_text)
+
+            # Store the string representation in the database
+            OrganizationSetting.upsert(
+                key=OrganizationSetting.DIRECTORY_INTRO,
+                value=str(sanitized_intro_text),  # Convert Markup to string
```
brassy-endomorph, line 837: We can't use `str(sanitized_tex)` because that will return an _escaped_ version of the string, so every time we do this, we will additoinally escape the text.

Try making the input the following string:

```
script -> <script>alert(1)</script>
hash -> #
amp -> &
```

Submit then refresh. Note that we have saved the escaped text to the DB which is not what we want.

`sanitize_input` should just return a `str` and we should manually wrap it for display like `Markup(sanitize_input(foo))` each time, but _not_ when sanitizing it here for DB insertion

tests/test_settings.py
```diff
@@ -666,3 +667,60 @@ def test_update_brand_logo(client: FlaskClient, admin: User) -> None:
     resp = client.get(logo_url, follow_redirects=True)
     # yes this check is ridiculous. why? because we redirect not-founds instead of actually 404-ing
     assert "That page doesn" in resp.text
+
+
+def test_sanitize_input() -> None:
+    from markupsafe import Markup
```
brassy-endomorph, line 673: import should be at top of file, not in the function

brassy-endomorph (changes requested):
Sorry, still some changes, one of them being my fault for pointing you a little in the wrong direction :/

hushline/routes.py
glenn-sorrentino: good catch. yes, looks like a dupe.

glenn-sorrentino: c28c55a

hushline/settings/__init__.py
glenn-sorrentino: got it. updated here: 5607dad

tests/test_settings.py
glenn-sorrentino: 5ad4d35

brassy-endomorph:
Noting that this was reverted because it got accidentally merged.

PR #767, opened 2024-11-26
message replies

### LINKED ISSUE #631: Message Replies
## Use Cases

- As a **tip line owner**, I need to reply to a whistleblower's message directly inside of Hush Line, so that the source knows the status of their disclosure.
- As a **whistleblower**, I need to know if my message has been received and its status, so that I know if I should look elsewhere to get help.

## Requirements

**Whistleblower**
- When an **unauthenticated** user submits a tip, a dynamic message URL is automatically created for them to the effect of `https://tips.hushline.app/to/Ungrouped-Headphones-Aluminum-9`
- A modal opens which gives the whistleblower a message containing their auto-generated URL.
- Clicking Continue redirects the whistleblower to the message page.
- The whistleblower only has access to this page. The global nav removes everything except the Directory and Donation links.

**Tip Line Owner**
- New messages receive an autogenerated name
- Drilling-in to a message shows the message metadata, encrypted contents, and a case status menu button. 
- Changing the case's status sends a predefined reply visible on the message page to both tip line owner and whistleblower.

### Whistleblower

**Step 1: Send a message**
When an unauthenticated user submits a new message, a modal opens displaying an autogen reply URL.

Image

**Step 2: Response Page**

Clicking Continue brings the user to the message page where they can see a copy of their message. A badge prominently displays the status.

Reply Page

**When a case has been accepted:**
Reply Page Accepted

### Tip Line Owner

**Step 1: New Message**

A new message is displayed with a blue indicator dot. 

Inbox - New Message

**Step 2: Message Status**

Drilling in to the message shows actions which allow the tip line owner to change its status. Replies are automatically sent when the case status changes.

Message - Waiting with Menu

brassy-endomorph:
- creates a per user per status markdown reply
- creates a per message reply page where submitters can see their message's status

TODOs

- it's ugly lol
- decide on what markdown/tags are supported
    - document it

fixes #631


glenn-sorrentino (changes requested):
Looking great. I updated a few things in #768.

glenn-sorrentino (approved):

PR #768, opened 2024-11-27
style tweaks

glenn-sorrentino:
1. CSS Updates:
	• Refactored .message, .message-container, and .user styles for consistent padding, borders, and box shadows.
	• Introduced .action-model styles for aligned and spaced action buttons/forms.
	• Improved styles for .message-meta and .message-header to enhance readability.
2. Template Enhancements:
	• Consolidated metadata display for messages in message.html and reply.html into reusable message-container structure.
	• Updated submission_success.html:
	• Removed redundant p tags around “Copy Link” button for cleaner layout.
	• Retained “Do not share this link” warning for emphasis.
3. Code Simplifications:
	• Reduced redundant markup in reply.html by aligning it with message.html.
	• Removed unnecessary styles and elements for better maintainability.
4. User Feedback Improvements:
	• Added .badge class to visually differentiate message status in reply.html.
	• Improved visual hierarchy in metadata and action sections.

Screenshot 2024-11-26 at 4 14 11 PM
Screenshot 2024-11-26 at 4 14 08 PM
Screenshot 2024-11-26 at 4 14 04 PM
Screenshot 2024-11-26 at 4 13 55 PM

### DIFF
--- a/hushline/templates/submission_success.html
+++ b/hushline/templates/submission_success.html
@@ -3,6 +3,7 @@
 {% block title %}Message Submitted{% endblock %}
 
 {% block content %}
+<h1>👍 Message Submitted!</h1>
   <p>
     Thank you for submitting your mesasge to {{ message.username.display_name or message.username.username }}!
     You should expect a response within a few days.
@@ -16,9 +17,7 @@
     {% endwith %}
   </p>
 
-  <p>
-    <a id="copy-link-button" class="btn" data-target="reply-url" href="#">Copy Link</a>
-  </p>
+  <a id="copy-link-button" class="btn" data-target="reply-url" href="#">Copy Link</a>
 
   <p>
     <strong>Do not</strong> share this linkt with anyone else.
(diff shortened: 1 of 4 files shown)


hushline/templates/submission_success.html
```diff
@@ -29,10 +28,6 @@
     Make sure you copy this link before closing this window or navigating to another page.
     Once you do, this link will not be displayed again.
   </p>
-
-  <p>
-    <a class="btn" href="{{ url_for('directory') }}">Continue</a>
```
brassy-endomorph, line 34: I think we still need a link back to "home" here somehow so the user has a clear exit. Maybe not a button, but at least a link like "Back to the directory," no?

glenn-sorrentino: I don't know, once the message has been sent I think the only action is exit, but adding a button for that just recreates browser behavior of closing the tab or window.

brassy-endomorph: hmm well we'll just leave this change for now an then figure it out later

Issue #663, opened 2024-10-08
404's should display an error page and not redirect


brassy-endomorph:
**Is your feature request related to a problem? Please describe.**

When a user access a resource that isn't found (not authorized, missing, whatever), they should land on an actual missing page that displays a generic "not found" error message whose status code is 404. No flashed message should be displayed.

**Why?**

We have some non-standard code that has already led me to confusion as to why `abort(404)` (a typical pattern) wasn't working as expected.

Affected endpoint:

https://github.com/scidsg/hushline/blob/4552cb0d4af6caaccacadc6bb47d1f1cf1b3e498/hushline/routes.py#L553-L562

Affected error handler:

https://github.com/scidsg/hushline/blob/4552cb0d4af6caaccacadc6bb47d1f1cf1b3e498/hushline/__init__.py#L95-L98

Also, this will make it possible for tests to check that a missing/unauthorized entity actually returns 404 (a clear sign of missing) instead of redirects to `/` and then doesn't have some heading (`User 123`) present in the response body.

PR #789, opened 2024-12-01
feat: add back button to settings/enable-2fa page to allow user to abort

### LINKED ISSUE #761: 2FA toggle page should have a "back" button so a user can abort the process
See the title

anGie44:
Closes #761 

## What  and How 🛠️ 

This PR introduces a back button in the `settings/enable-2fa` view. The button is added via the `<a>` element in the page's template file.

## Why 🤷 

The new link serves as a an escape hatch, allowing users to return to the authentication settings page.

## Testing 🧪 

To validate the existence and usage of the `<a>` element representing the back button, I've leveraged [Playwright](https://playwright.dev/python/docs/intro)'s browser testing. The test cases live in a `ui` subdirectory so they can be run independently with a new target, namely `make test/ui`, or alongside all the the tests via `make test`.

Locally, you can pull this branch and run `make test/ui` to validate.

## Anything else 💭 

I chose Playwright here instead of Selenium because it seems to spin up faster and perhaps less overhead -- did require installing some dependencies (see `Dockerfile.dev`).  Open to discussing this choice more since sets a precedent for e2e browser testing

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -21,7 +21,11 @@ RUN apt-get update && \
 RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
 ENV PATH="/root/.cargo/bin:${PATH}"
 
-RUN pip install poetry
+RUN pip install poetry pytest-playwright
+
+# Playwright dependencies to run browsers
+RUN playwright install
+RUN playwright install-deps
 
 WORKDIR /app
 
--- a/Makefile
+++ b/Makefile
@@ -50,8 +50,13 @@ ifndef message
 endif
 	poetry run flask db revision -m "$(message)" --autogenerate
 
-TESTS ?= ./tests/
+TESTS ?= ./tests
 .PHONY: test
 test: ## Run the test suite
 	docker compose run --rm app \
 		poetry run pytest --cov hushline --cov-report term --cov-report html -vv $(PYTEST_ADDOPTS) $(TESTS)
+
+.PHONY: test/ui
+test/ui: # Only run the ui tests
+	docker compose run --rm app \
+		poetry run pytest --cov hushline --cov-report term --cov-report html -vv $(PYTEST_ADDOPTS) $(TESTS)/ui
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,6 +29,8 @@ stripe = "^10.9.0"
 types-requests = "^2.32.0.20240712"
 types-setuptools = "^71.1.0.20240813"
 python = "^3.11"
+pytest-playwright = "^0.6.2"
+tenacity = "^9.0.0"
 
 [tool.poetry.group.dev.dependencies]
 beautifulsoup4 = "^4.12.3"
--- a/tests/ui/test_settings_enable_2fa.py
+++ b/tests/ui/test_settings_enable_2fa.py
@@ -1,0 +1,43 @@
+import re
+
+from playwright.sync_api import Page, expect
+from tenacity import RetryCallState, retry, stop_after_attempt, wait_random_exponential
+
+BASE_URL = "http://127.0.0.1:8080"
+
+
+def log_retry_error(retry_state: RetryCallState) -> None:
+    if retry_state is None or retry_state.outcome is None:
+        return
+    print(f"Retrying due to error: {retry_state.outcome.exception()}")
+
+
+@retry(
+    wait=wait_random_exponential(min=1, max=60),
+    stop=stop_after_attempt(3),
+    retry_error_callback=log_retry_error,
+)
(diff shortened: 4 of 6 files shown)


tests/ui/test_settings_enable_2fa.py
```diff
@@ -0,0 +1,43 @@
...
+BASE_URL = "http://127.0.0.1:8080"
+
+
+def log_retry_error(retry_state: RetryCallState) -> None:
+    if retry_state is None or retry_state.outcome is None:
+        return
+    print(f"Retrying due to error: {retry_state.outcome.exception()}")
+
+
+@retry(
```
anGie44, line 15: Required retrying initial loading of `localhost:8080/login` because it seems the playwright's browser can spin up before the app is ready to serve requests.

tests/ui/test_settings_enable_2fa.py
```diff
@@ -0,0 +1,43 @@
...
+    wait=wait_random_exponential(min=1, max=60),
+    stop=stop_after_attempt(3),
+    retry_error_callback=log_retry_error,
+)
+def test_enable_2fa_has_back_button(page: Page, user_password: str) -> None:
+    page.goto(f"{BASE_URL}/login", wait_until="domcontentloaded")
+    page.get_by_label("Username").fill("test")
+    page.get_by_label("Password").fill(user_password)
+    page.get_by_role("button", name="Login").click()
```
anGie44, line 24: i tried moving this out into a `pytest` fixture for the test methods to use before testing the actual button-related changes but couldn't get the tests to synchronously wait for a successful login 🤔 so there's duplication here and test below which isn't great but in essence of time, could still be a good first pass at integrating some browser testing.

Dockerfile.dev
```diff
@@ -21,7 +21,11 @@ RUN apt-get update && \
 RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
 ENV PATH="/root/.cargo/bin:${PATH}"
 
-RUN pip install poetry
+RUN pip install poetry pytest-playwright
+
+# Playwright dependencies to run browsers
+RUN playwright install
```
brassy-endomorph, line 27: Shouldn't these go after the `poetry install` so that we get the versions that are pinned in the `poetry.lock` file? Is there an equivalent `playwritght` type lockfile

Makefile
```diff
@@ -50,8 +50,13 @@ ifndef message
...
 
-TESTS ?= ./tests/
+TESTS ?= ./tests
 .PHONY: test
 test: ## Run the test suite
 	docker compose run --rm app \
 		poetry run pytest --cov hushline --cov-report term --cov-report html -vv $(PYTEST_ADDOPTS) $(TESTS)
+
+.PHONY: test/ui
+test/ui: # Only run the ui tests
```
brassy-endomorph, line 60: I think we way we'll want to split this is either using some kind of `pytest.mark` or possibly a wholly separate directory like `ui-test` since we'd need a separate set of fixtures for a live app, though there might be some clean way to make them overlap.

pyproject.toml
```diff
@@ -29,6 +29,8 @@ stripe = "^10.9.0"
 types-requests = "^2.32.0.20240712"
 types-setuptools = "^71.1.0.20240813"
 python = "^3.11"
+pytest-playwright = "^0.6.2"
```
brassy-endomorph, line 32: nit: alphabetize

tests/ui/test_settings_enable_2fa.py
```diff
@@ -0,0 +1,43 @@
+import re
+
+from playwright.sync_api import Page, expect
+from tenacity import RetryCallState, retry, stop_after_attempt, wait_random_exponential
+
+BASE_URL = "http://127.0.0.1:8080"
```
brassy-endomorph, line 6: This URL (or rather, port) should probably come via a fixture of some sort using a trick like this:

https://github.com/freedomofpress/securedrop/blob/6ad1fa95330c4f36c3144cb871620c64d8bc1f94/securedrop/tests/functional/conftest.py#L43-L48

I'm guessing the UI tests require an already running app that's started in a second terminal or so, yeah?

tests/ui/test_settings_enable_2fa.py
brassy-endomorph: If we use pytest fixtures, we can have a liveness check on the app so that it doesn't even return/yield said fixture until it's usable.

brassy-endomorph (commented):
I think we need to collaborate on a better UI testing strategy since that itself is no small task. There's also so cleanup in the app code like #663 that would make it much easier to test. I would say that likely using a live app + selenium-eque driver would only be for things like JS-only features due to the complexities around setting up an app like that in a way that preserves fixture isolation.

PR #805, opened 2024-12-08
Minor: Add min-width to buttons

glenn-sorrentino:
This allows for pagination dots to be centered when the exit button's label is "exit."


scidsg-dev (approved):

scidsg-dev (approved):

PR #806, opened 2024-12-09
fernet encrypted sessions

### LINKED ISSUE #603: Sessions are not encrypted and leak `user_id`

Flask sessions are not encrypted, only signed/HMAC'd. This leaks user counts.

Ideally both, but at least one of:

- encrypt cookies
- use UUIDs for user IDs

brassy-endomorph:
Overrides the default session interface to use on that is encrypted and timed via Fernet. The logic largely copies the internal logic of Flask's default `SecureSessionInterface`, so comparison should be quite easy.

Flask Docs: https://flask.palletsprojects.com/en/3.0.x/api/#flask.sessions.SessionInterface

Fixes #603


micahflee (approved):
Nice, I just tested this and it works great. I even manually decrypted the session cookie using the python interpreter to make sure,

PR #843, opened 2024-12-20
Fix Extra Fields Population

### LINKED ISSUE #842: Saved Extra Field Values are Not Displaying in the UI
Extra fields value are not being displayed after being saved.

glenn-sorrentino:
This fixes #842 - a bug where extra field values are not displayed after being saved.

- The original `ProfileForm` initialization `(obj=user)` is replaced with a manual assignment of bio and dynamically populated `extra_field_label{I}` and `extra_field_value{I}` fields.
- Each extra field uses `getattr(username, ...)` to fetch its saved value.


hushline/settings/__init__.py
```diff
@@ -329,15 +329,24 @@ async def profile() -> Response | Tuple[str, int]:
         user = db.session.scalars(db.select(User).filter_by(id=session["user_id"])).one()
         username = user.primary_username
 
-        # hush, mypy.
```
brassy-endomorph, line 332: :((((

brassy-endomorph (approved):

PR #855, opened 2025-01-06
Registration CAPTCHA

### LINKED ISSUE #854: Add Math CAPTCHA to Registration Flow

glenn-sorrentino:
Closes #854 

This PR introduces CAPTCHA validation to the registration process to enhance security and prevent automated spam registrations. The implementation mirrors the existing CAPTCHA logic used in the message submission route, ensuring consistency and simplicity.

## Key Changes
1.	Routes Update (hushline/routes.py):
•	Added a math CAPTCHA to the /register route:
•	Generates a math problem on GET or when no existing CAPTCHA is in the session.
•	Reuses the same CAPTCHA on failed submissions for retry.
•	Refactored logic to ensure consistent handling of CAPTCHA across form validation.
2.	Template Update (hushline/templates/register.html):
•	Added a CAPTCHA input field for users to solve during registration.
•	Displays the generated math problem (math_problem) dynamically.
3.	Version Update (hushline/version.py):
•	Incremented the version from 0.3.68 to 0.3.69.
4.	Bug Fixes and Enhancements:
•	Simplified the /register route’s return type to Response | str (removed unnecessary tuple return type).
•	Fixed status code handling for invalid invite codes and existing usernames.

## Additional Notes
•	The CAPTCHA logic was adapted directly from the /to/<username> route for consistency.
•	This change impacts only the registration flow and does not affect existing routes or logic.


micahflee (approved):
Looks good!

I don't think it's necessary for this PR, but since we now have very similar code for generating math CAPTCHAs in multiple places, it might be good to separate this into its own function/set of functions, and use those functions. This will make it easier to plot CAPTCHAs wherever we want. Or if we ever want to change the type of CAPTCHA we use, we can change it one place.

PR #881, opened 2025-01-23
Remove REQUIRE_PGP env var and instead always require PGP

micahflee:
Please double check the logic changes in `hushline/templates/profile.html`.


brassy-endomorph (commented):
needs a docs update

brassy-endomorph (approved):

PR #905, opened 2025-02-06
Make additional fields for paid/single tentant users

micahflee:
Similar to `ALIAS_MODE`, this adds a new environment variable `FIELDS_MODE` which can either be `always` (the default) or `premium`.

I added a `User.fields_enabled` property which determines if fields are enabled for this user. If `FIELDS_MODE` is `always` it returns True. Otherwise, it only returns True if the user has a premium account.

If `User.fields_enabled` is False, the templates hide the fields-updating code, and the server-side routes return 401.


hushline/model/user.py
```diff
@@ -179,6 +179,15 @@ def max_aliases(self) -> int:
         current_app.logger.warning(err_msg)
         return self._PREMIUM_ALIAS_COUNT
 
+    @property
```
brassy-endomorph, line 182: very much like that this is its own property. more understandable, easier to refactor :D

brassy-endomorph (commented):
Would be nice to add tests for the form update endpoint to ensure it 401s.

Also this is relevant to this as much as it is to aliases, but we don't degrade gracefully from pro to free on either if the master config var or the payment tier change. Not saying we have to solve this here, but we do need to consider solving it soon.

glenn-sorrentino (dismissed):
Behavior looks great

glenn-sorrentino (approved):

PR #918, opened 2025-02-13
Docs Update

glenn-sorrentino:
Comes into parity with hushline.app/getting-started.html


scidsg-dev (dismissed):

PR #927, opened 2025-02-15
Fix verification with explicit HTTPS check 

glenn-sorrentino:


brassy-endomorph (changes requested):
please do not merge this as this is almost certainly not the right fix as we should be looking at `PREFERRED_URL_SCHEME` and figuring out 1) if we're setting it and 2) why it's not working

PR #936, opened 2025-02-20
Option to encrypt entire email body

### LINKED ISSUE #908: Message Formatting Option
**Problem**
Currently messages have individual fields encrypted, causing issues with automatic decryption in clients like Proton. 

**Solution**
Add an additional toggle in Settings > Notifications:

```
[-o] Tip Notifications
[-o] Include Message Contents
[-o] Compatible Message Formatting <--NEW

```

micahflee:
Fixes #908.

This adds an "Encrypt Entire Body" setting:

image

When submitting the submission form, in addition to encrypting each encrypted field separately, it also concatenates all of the fields (encrypted or not) together as a single email body and encrypts that too, and saves it in a hidden field.

If the user has email notifications enabled, and include message content is enabled, and encrypt entire body is enabled, it sends this encrypted body. If for some reason the client-side encryption broke, it falls back to how it currently works.


glenn-sorrentino (dismissed):

hushline/routes/forms.py
```diff
@@ -62,6 +63,15 @@ class F(FlaskForm):
 
         self.F = F
 
+        # Add email body hidden field
+        setattr(
```
brassy-endomorph, line 67: You can just do this like a normal `FlaskForm`  in the class definition for `F`. Not with `setattr`

hushline/routes/profile.py
```diff
@@ -119,12 +119,21 @@ def submit_message(username: str) -> Response | str:
             db.session.commit()
 
             if uname.user.enable_email_notifications:
-                if not uname.user.email_include_message_content:
-                    email_body = "You have a new Hush Line message! Please log in to read it."
+                if uname.user.email_include_message_content:
```
brassy-endomorph, line 122: We don't have a test for this fairly critical path. Also might want to pessimistically check the body is encrypted on the backend and reencrypt.

micahflee: Re-encrypting on the backend won't work if the form has any encrypted fields, because then they'll be double-encrypted, and so won't be readable in the email client. Adding a test is a good idea though

brassy-endomorph: Maybe at least test that the PGP start string exists at all in the message and then encrypt. Maybe that's annoying UX in some edge cases, but it would otherwise leak data which is worse imo.

hushline/routes/forms.py
micahflee: Good point

hushline/routes/profile.py
micahflee: Instead of re-encrypting, I've changed it so that if `encrypted_email_body` is not PGP-encrypted, it falls back to the generic "You have a new Hush Line message" email

micahflee: And I wrote notifications tests!

glenn-sorrentino (approved):

PR #943, opened 2025-02-28
Add FIELDS_MODE to docs

micahflee:


scidsg-dev (approved):

PR #960, opened 2025-03-10
Inbox fit and finish

glenn-sorrentino:
Fixes scrolling on empty states.


scidsg-dev (dismissed):

Issue #975, opened 2025-03-29
Branding missing from Personal Server settings


glenn-sorrentino:
Since this is all unlocked, branding should be visible to the end user. I think our logic is tied to stripe currently.

brassy-endomorph:
Nvm, answer in this comment: https://github.com/scidsg/hushline/issues/972#issuecomment-2768108551

PR #979, opened 2025-04-03
First user automatically becomes an admin

### LINKED ISSUE #972: First user should automatically become an admin
When registering an account, if there are no users yet, the first user should be an admin user. Otherwise, users should not be admins. This is important for the personal server.

micahflee:
Fixes #972 and #975

I've added a new docker compose file, `docker-compose.personal-server.yaml`, to simulate the personal server. Specifically, I needed to be able to test this in dev mode but without dev data. You can test with:

```sh
docker compose -f docker-compose.personal-server.yaml up --build
```

This does a few things:

When loading the index, if there are no users, it redirects to the register page.

If there are no users, the register page shows a special alert to create the first admin user:

Screenshot from 2025-04-03 13-48-05

For the styles, I decided to make the alert look just like the upgrade buttons in settings. To do this, I renamed the `upgrade` class to `alert` so we can re-use this style in other places.

When creating the first user, it make the user an admin, and it also by default adds the user's primary username to the directory.

I have tests for all of this.

---

Now that Personal Server users can be admins, I discovered problems we should fix in a separate issue:

- Settings > Admin: You can toggle verified users. But I think it makes sense to remove verified functionality altogether except from the hosted service.
- Settings > Admin: You can make yourself no longer an admin. If you click this, it happens immediately, and now the personal server has no admins and the user is locked out forever. At the very least, we need a check that if the user is the only admin, they can't remove admin from themselves.
- Settings > Advanced: you can delete the only admin account. That should also be disallowed.

### DIFF
--- a/docker-compose.personal-server.yaml
+++ b/docker-compose.personal-server.yaml
@@ -1,0 +1,47 @@
+---
+services:
+  app:
+    build:
+      context: .
+      dockerfile: Dockerfile.prod
+    ports:
+      - 127.0.0.1:8080:8080
+    environment:
+      IS_DOCKER: '1'
+      FLASK_APP: hushline
+      FLASK_ENV: production
+      ENCRYPTION_KEY: bi5FDwhZGKfc4urLJ_ChGtIAaOPgxd3RDOhnvct10mw=
+      SECRET_KEY: cb3f4afde364bfb3956b97ca22ef4d2b593d9d980a4330686267cabcd2c0befd
+      SESSION_FERNET_KEY: jY0gDbATEOQolx2SGj46YnkkbN6HQBB4YCABzwl1H1A=
+      SQLALCHEMY_DATABASE_URI: postgresql://hushline:hushline@postgres:5432/hushline
+      REGISTRATION_CODES_REQUIRED: "false"
+      ALIAS_MODE: always
+      FIELDS_MODE: always
+      SESSION_COOKIE_NAME: session
+      DIRECTORY_VERIFIED_TAB_ENABLED: "false"
+      BLOB_STORAGE_PUBLIC_DRIVER: file-system
+      BLOB_STORAGE_PUBLIC_FS_ROOT: /hushline-public-files
+    depends_on:
+      postgres:
+        condition: service_healthy
+    volumes:
+      - hushline-public-files:/hushline-public-files
+    restart: always
+
+  postgres:
+    image: postgres:16.4-alpine3.20
+    environment:
+      POSTGRES_USER: hushline
+      POSTGRES_PASSWORD: hushline
+      POSTGRES_DB: hushline
+    restart: on-failure
+    ports:
+      - 127.0.0.1:5432:5432
+    healthcheck:
+      test: ["CMD", "pg_isready", "-U", "hushline"]
+      interval: 5s
+      timeout: 1s
+      retries: 10
+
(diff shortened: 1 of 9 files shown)


glenn-sorrentino (approved):

glenn-sorrentino (approved):

docker-compose.personal-server.yaml
```diff
@@ -0,0 +1,57 @@
...
+    environment:
+      IS_DOCKER: '1'
+      FLASK_APP: hushline
+      FLASK_ENV: development
+      ENCRYPTION_KEY: bi5FDwhZGKfc4urLJ_ChGtIAaOPgxd3RDOhnvct10mw=
+      SECRET_KEY: cb3f4afde364bfb3956b97ca22ef4d2b593d9d980a4330686267cabcd2c0befd
+      SESSION_FERNET_KEY: jY0gDbATEOQolx2SGj46YnkkbN6HQBB4YCABzwl1H1A=
+      SQLALCHEMY_DATABASE_URI: postgresql://hushline:hushline@postgres:5432/hushline
+      REGISTRATION_CODES_REQUIRED: "false"
+      ALIAS_MODE: always
```
brassy-endomorph, line 18: Maybe we should have these as make targets and toggle the few env vars. Would that be sufficient?

```yaml
  FOO_VAR: ${FOO_VAR}
```

hushline/routes/auth.py
```diff
@@ -51,6 +51,9 @@ def register() -> Response | str:
             # Use the existing math problem from the session
             math_problem = session.get("math_problem", "Error: CAPTCHA not generated.")
 
+        # Check if this is the first user
+        first_user = db.session.query(User).count() == 0
```
brassy-endomorph, line 55: TOCTOU error here. This could (should?) be done in an explicit transaction.

tests/helpers.py
```diff
@@ -61,6 +61,18 @@ def __ne__(self, other: object) -> bool:
 assert Missing() != Missing()
 
 
+def get_captcha_from_session_register(client: FlaskClient) -> str:
```
brassy-endomorph, line 64: Could just speed up tests and set `math_answer` to a hard coded variable for testing and then just always set that in requests to that endpoint. Or, could just disable CAPTCHAs for testing.

tests/test_first_user.py
```diff
@@ -0,0 +1,104 @@
...
+    assert response.status_code == 200
+
+    assert "Create the Admin User" not in response.text
+
+
+def test_first_user_is_admin(client: FlaskClient) -> None:
+    user_count = db.session.query(User).count()
+    assert user_count == 0
+
+    os.environ["REGISTRATION_CODES_REQUIRED"] = "False"
```
brassy-endomorph, line 50: Should use a mock fixture or decorator for this

tests/test_first_user.py
```diff
@@ -0,0 +1,104 @@
...
+    assert user_count == 0
+
+    user = User(password=user_password)
+    user.tier_id = 1
+    user.is_admin = True
+    db.session.add(user)
+    db.session.commit()
+    assert db.session.query(User).count() == 1
+
+    os.environ["REGISTRATION_CODES_REQUIRED"] = "False"
```
brassy-endomorph, line 84: mock fixture/decorator

PR #986, opened 2025-04-10
Header CSS

glenn-sorrentino:
- Adds a fixed-height to the header.
- Updates logo line height.


scidsg-dev (approved):

PR #1007, opened 2025-04-18
Prevent the only admin user from deleting itself

### LINKED ISSUE #981: Prevent admins from removing the only admin
Right now there are two ways that admins can remove their own admin access:

- Settings > Admin: You can remove admin access from your own user.
- Settings > Advanced: You can delete your account.

If you're logged in as an admin, and if you are the only admin user, both of these should be disabled. You shouldn't be able to remove admin access from the only admin (there should be at least one). And you shouldn't be able to delete the only admin account.

Related to #972.

micahflee:
Fixes #981.

This PR is based on #1006 (because I extend the tests) so that should be reviewed and merged first.


glenn-sorrentino (approved):

hushline/admin.py
```diff
@@ -27,6 +30,13 @@ def toggle_admin(user_id: int) -> Response:
         user = db.session.get(User, user_id)
         if user is None:
             abort(404)
+
+        if user.is_admin:
```
brassy-endomorph, line 34: TOCTOU error with race condition here. This should be done in an explicit transaction.

hushline/settings/delete_account.py
```diff
@@ -31,6 +32,12 @@ def delete_account() -> Response | str:
 
         user = db.session.get(User, user_id)
         if user:
+            if user.is_admin:
```
brassy-endomorph: TOCTOU error again. Should be in a single transaction.

brassy-endomorph, line 35: TOCTOU error again. Wrap with transaction.

tests/test_delete_account.py
```diff
@@ -0,0 +1,63 @@
...
+from flask.testing import FlaskClient
+
+from hushline.db import db
+from hushline.model import (
+    User,
+)
+
+
+@pytest.mark.usefixtures("_authenticated_user")
+def test_delete_account(client: FlaskClient, user: User) -> None:
```
brassy-endomorph, line 12: Being able to delete the last user even when not an admin feels wrong. Is this the behavior we want?

brassy-endomorph (commented):
Fine as is, but we might want to handle the edge cases.

brassy-endomorph (approved):

Issue #718, opened 2024-10-29
Don't have flashed messages disappear automatically


brassy-endomorph:
**Is your feature request related to a problem? Please describe.**

Flashed messages disappear after... 5 seconds? Something short. This is not great for accesibility as important information might be lost either because of inattention, distraction, or some other disability. Heck, even ESL.

**Describe the solution you'd like**

They should have an `X` to close them if we want them to go away.

PR #1010, opened 2025-04-23
Registration settings

### LINKED ISSUE #973: Better registration settings
The personal server has REGISTRATION_CODES_REQUIRED set to false and this is not editable, which means anyone can register accounts on any personal server. Some people might want to disable this.

I think we should define some global settings (stored in the database) for:

- enable registration (default true), so admins can choose to just disable all registration after they've made their one account
- registration codes required, so admins can change this setting, if registration is enabled

And we should also have a user interface to manage registration codes. Right now it's done with a script you run in the terminal, but admins should be able to generate and delete registration codes from the settings.

micahflee:
Fixes #973.

There's now a registration settings page, only accessible to admins:

image

You can turn on registration, required registration codes, and also create and delete codes:

image

image

image

If registrations are disabled, the `/register` route redirects to index, and the Register link doesn't get displayed in the nav bar.

This also fixes two small bugs:

- The `register.html` template needs a variable `first_user` and it was only being passed in on GET requests, not POST requests, so this fixes that
- Invite codes are deleted after use

This also completely deletes the REGISTRATION_CODES_REQUIRED environment variable. By default this setting was disabled, and I doubt it's actively used in the wild (since there's no interface for generating invite codes). So this just means that if there is an instance that has this env var set to true, this PR will make it so they're no longer required, and the user will need to login as an admin and change the setting there. (The alternative would to write a little migration, which doesn't seem worth it since I don't think it's being used.)

### DIFF
--- a/assets/scss/style.scss
+++ b/assets/scss/style.scss
@@ -3428,3 +3428,30 @@ li.tab .badge {
   text-align: center;
   padding: 1rem;
 }
+
+.registration-code-list {
+  display: flex;
+  flex-direction: column;
+  gap: 1rem;
+  margin-top: 1rem;
+}
+
+.registration-code-entry {
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  align-items: center;
+  position: relative;
+  /* needed for stretched-link */
+}
+
+.registration-code-entry .registration-code {
+  font-family: var(--font-sans-bold);
+  font-size: var(--font-size-base);
+}
+
+.registration-code-entry .registration-code-exp {
+  font-family: var(--font-sans);
+  font-size: var(--font-size-small);
+  color: var(--color-gray);
+}
--- a/hushline/__init__.py
+++ b/hushline/__init__.py
@@ -132,6 +132,8 @@ def configure_jinja(app: Flask) -> None:
             OrganizationSetting.GUIDANCE_EXIT_BUTTON_LINK,
             OrganizationSetting.GUIDANCE_PROMPTS,
             OrganizationSetting.HIDE_DONATE_BUTTON,
+            OrganizationSetting.REGISTRATION_ENABLED,
+            OrganizationSetting.REGISTRATION_CODES_REQUIRED,
         )
 
         data.update(
@@ -140,6 +142,10 @@ def configure_jinja(app: Flask) -> None:
             directory_verified_tab_enabled=app.config["DIRECTORY_VERIFIED_TAB_ENABLED"],
             is_onion_service=request.host.lower().endswith(".onion"),
             is_premium_enabled=bool(app.config.get("STRIPE_SECRET_KEY", False)),
+            registration_enabled=data.get(OrganizationSetting.REGISTRATION_ENABLED, True),
+            registration_codes_required=data.get(
+                OrganizationSetting.REGISTRATION_CODES_REQUIRED, False
+            ),
         )
 
         if "user_id" in session:
--- a/hushline/settings/registration.py
+++ b/hushline/settings/registration.py
@@ -3,13 +3,14 @@ from typing import Tuple
 from flask import Blueprint, flash, redirect, render_template, request, url_for
 from flask_wtf import FlaskForm
 from werkzeug.wrappers.response import Response
-from wtforms import BooleanField, SubmitField
+from wtforms import BooleanField, IntegerField, SubmitField
+from wtforms.validators import DataRequired
 from wtforms.validators import Optional as OptionalField
 
 from hushline.auth import admin_authentication_required
 from hushline.db import db
-from hushline.forms import DisplayNoneButton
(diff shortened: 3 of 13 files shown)


glenn-sorrentino (changes requested):
If we're removing `REGISTRATION_CODES_REQUIRED` how will we gate licensed managed single-tenant instances?

micahflee:
Before this PR is ready for review, we need to add a new separate CLI interface for managing registration. It should work like this:

- There should be a new environment variable DISABLE_REGISTRATION_SETTINGS, default to False
- If it's True, then the registration settings in this PR are all is disabled
- There should be an additional CLI interface to manage registrations that's always enabled with the ability to:
  - Enabled/disable registrations
  - Enable/disable requiring registration codes
  - List/create/delete registration codes

assets/scss/style.scss
```diff
@@ -3428,3 +3428,30 @@ li.tab .badge {
   text-align: center;
   padding: 1rem;
 }
+
+.registration-code-list {
```
brassy-endomorph, line 3432: Could you split this into its own file called like `assets/scss/settings/registration.scss` so we can eventually split this massive unsorted file into something more modular? Then include it at the top of this one with [`@forward`](https://sass-lang.com/documentation/at-rules/forward/).

hushline/__init__.py
```diff
@@ -140,6 +142,10 @@ def inject_variables() -> dict[str, Any]:
             directory_verified_tab_enabled=app.config["DIRECTORY_VERIFIED_TAB_ENABLED"],
             is_onion_service=request.host.lower().endswith(".onion"),
             is_premium_enabled=bool(app.config.get("STRIPE_SECRET_KEY", False)),
+            registration_enabled=data.get(OrganizationSetting.REGISTRATION_ENABLED, True),
```
brassy-endomorph, line 145: For security, we might want to make this false by default, no?

hushline/settings/registration.py
```diff
@@ -0,0 +1,131 @@
...
+                return redirect(url_for(".registration"))
+            # Create invite code
+            elif (
+                create_invite_code_form.submit.name in request.form
+                and create_invite_code_form.validate()
+            ):
+                new_invite_code = InviteCode()
+                db.session.add(new_invite_code)
+                db.session.commit()
+                flash(f"Invite code {new_invite_code.code} created.")
```
brassy-endomorph, line 87: As a note, our flashes disappear after like 2 seconds (see #718) and this is going to mean users loose their registration codes quite quick after creation.

brassy-endomorph (commented):
Minor comments

hushline/__init__.py
micahflee: The problem is registration needs to be enabled by default to register the first user on the personal server.

A solution would be to set REGISTRATION_ENABLED to False by default, but in practice it's enabled until there is at least 1 user. Thoughts @brassy-endomorph ?

hushline/settings/registration.py
micahflee: It also displays the registration code in a table on the same page it's loading, so the flash message isn't actually all that important here.

micahflee:
I have addressed @brassy-endomorph's comments, including making REGISTRATION_ENABLED default to False, but with an exception to allow the first user to register if there are no users.

This also now has a CLI interface for managing registration settings and invite codes. To test it, start the server like this:

```sh
docker compose -f docker-compose.personal-server.yaml up --build
```

All registration commands start with `poetry run flask reg`. In another window, you can run:

```
$ docker compose exec app poetry run flask reg
Usage: flask reg [OPTIONS] COMMAND [ARGS]...

  Registration settings commands

Options:
  --help  Show this message and exit.

Commands:
  code-create                  Create an invite code
  code-delete                  Delete an invite code
  code-list                    List all invite codes
  registration-codes-required  Set REGISTRATION_CODES_REQUIRED to the...
  registration-enabled         Set REGISTRATION_ENABLED to the given value
  settings                     View registration settings
```

And finally, this adds a new environment variable, REGISTRATION_SETTINGS_ENABLED, which defaults to true. If it's false, registration settings don't appear in the settings page nav, and trying to access them anyway returns 401 errors.

glenn-sorrentino (dismissed):

hushline/cli_stripe.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def register_stripe_commands(app: Flask) -> None:
+    stripe_cli = AppGroup("stripe", help="Stripe commands")
+
+    @stripe_cli.command("configure")
+    def configure() -> None:
+        """Configure Stripe and premium tiers"""
+        # Make sure tiers exist
+        with app.app_context():
```
brassy-endomorph, line 18: App contexts also in this file

brassy-endomorph (dismissed):

glenn-sorrentino (approved):

PR #1011, opened 2025-04-23
rename MANAGED_SERVICE to USER_VERIFICATION_ENABLED

brassy-endomorph:


docs/README.md
```diff
@@ -406,6 +399,13 @@ These configs are needed for the web app.
       <td></td>
       <td><a href="https://docs.stripe.com/webhooks#endpoint-secrets" target="_blank">Stripe webhook key</a> for enabling payments</td>
     </tr>
+    <tr>
+      <td><code>USER_VERIFICATION_ENABLED</code></td>
+      <td>false</td>
+      <td>boolean</td>
+      <td><code>true</code></td>
```
micahflee, line 406: Should default to false, not true (we only want this feature on the main managed service)

hushline/config.py
```diff
@@ -143,7 +143,7 @@ def _load_hushline_misc(env: Mapping[str, str]) -> Mapping[str, Any]:
         ("DIRECTORY_VERIFIED_TAB_ENABLED", True),
         ("FILE_UPLOADS_ENABLED", False),
         ("REGISTRATION_CODES_REQUIRED", True),
-        ("MANAGED_SERVICE", False),
+        ("USER_VERIFICATION_ENABLED", True),
```
micahflee, line 146: Should default to false

micahflee (changes requested):
Just need to change the default from true to false. Otherwise it seems to work great.

glenn-sorrentino (approved):

PR #1017, opened 2025-04-29
Update models to match the current state of the db

### LINKED ISSUE #870: Python DB schema and migrations are out of sync
Running the migration script produces this migration when none should be produced at the `HEAD` of `main`

```python
def upgrade() -> None:
    with op.batch_alter_table('messages', schema=None) as batch_op:
        batch_op.alter_column('created_at',
               existing_type=postgresql.TIMESTAMP(timezone=True),
               type_=sa.DateTime(),
               existing_nullable=False,
               existing_server_default=sa.text('now()'))


def downgrade() -> None:
    with op.batch_alter_table('messages', schema=None) as batch_op:
        batch_op.alter_column('created_at',
               existing_type=sa.DateTime(),
               type_=postgresql.TIMESTAMP(timezone=True),
               existing_nullable=False,
               existing_server_default=sa.text('now()'))
```

micahflee:
Fixes #870 

Now when you create an autogenerated alembic migration, it doesn't create a migration at all.

It still has a few warnings though which I'm not sure how to get rid of, but I think we can safely ignore:

```
INFO  [alembic.ddl.postgresql] Detected sequence named 'message_id_seq' as owned by integer column 'messages(id)', assuming SERIAL and omitting
INFO  [alembic.ddl.postgresql] Detected sequence named 'invite_code_id_seq' as owned by integer column 'invite_codes(id)', assuming SERIAL and omitting
INFO  [alembic.ddl.postgresql] Detected sequence named 'authentication_logs_id_seq' as owned by integer column 'authentication_logs(id)', assuming SERIAL and omitting
INFO  [alembic.ddl.postgresql] Detected sequence named 'field_definitions_id_seq' as owned by integer column 'field_definitions(id)', assuming SERIAL and omitting
INFO  [alembic.ddl.postgresql] Detected sequence named 'message_status_text_id_seq' as owned by integer column 'message_status_text(id)', assuming SERIAL and omitting
INFO  [alembic.ddl.postgresql] Detected sequence named 'stripe_events_id_seq' as owned by integer column 'stripe_events(id)', assuming SERIAL and omitting
INFO  [alembic.ddl.postgresql] Detected sequence named 'field_values_id_seq' as owned by integer column 'field_values(id)', assuming SERIAL and omitting
INFO  [alembic.ddl.postgresql] Detected sequence named 'stripe_invoices_id_seq' as owned by integer column 'stripe_invoices(id)', assuming SERIAL and omitting
INFO  [alembic.env] No changes in schema detected.

```

### DIFF
--- a/hushline/model/field_definition.py
+++ b/hushline/model/field_definition.py
@@ -20,16 +20,20 @@ class FieldDefinition(Model):
     __tablename__ = "field_definitions"
     __table_args__ = (UniqueConstraint("username_id", "sort_order"),)
 
-    id: Mapped[int] = mapped_column(primary_key=True)
-    username_id: Mapped[int] = mapped_column(db.ForeignKey("usernames.id"), index=True)
+    id: Mapped[int] = mapped_column(primary_key=True, nullable=False)
+    username_id: Mapped[int] = mapped_column(
+        db.ForeignKey("usernames.id"), nullable=True, index=True
+    )
     username: Mapped["Username"] = relationship(back_populates="message_fields")
-    label: Mapped[str] = mapped_column(db.String(255))
-    field_type: Mapped[FieldType] = mapped_column(SQLAlchemyEnum(FieldType))
-    required: Mapped[bool] = mapped_column()
-    enabled: Mapped[bool] = mapped_column()
-    encrypted: Mapped[bool] = mapped_column()
-    choices: Mapped[list[str]] = mapped_column(type_=JSONB)
-    sort_order: Mapped[int] = mapped_column()
+    label: Mapped[str] = mapped_column(db.String(255), nullable=False)
+    field_type: Mapped[FieldType] = mapped_column(
+        SQLAlchemyEnum(FieldType, name="fieldtype"), nullable=True
+    )
+    required: Mapped[bool] = mapped_column(nullable=True)
+    enabled: Mapped[bool] = mapped_column(nullable=True)
+    encrypted: Mapped[bool] = mapped_column(nullable=True)
+    choices: Mapped[list[str]] = mapped_column(type_=JSONB, nullable=True)
+    sort_order: Mapped[int] = mapped_column(nullable=True)
 
     def __init__(  # noqa: PLR0913
         self,
--- a/hushline/model/field_definition.py
+++ b/hushline/model/field_definition.py
@@ -20,7 +20,7 @@ class FieldDefinition(Model):
     __tablename__ = "field_definitions"
     __table_args__ = (UniqueConstraint("username_id", "sort_order"),)
 
-    id: Mapped[int] = mapped_column(primary_key=True, nullable=False)
+    id: Mapped[int] = mapped_column(primary_key=True, nullable=False, autoincrement=True)
     username_id: Mapped[int] = mapped_column(
         db.ForeignKey("usernames.id"), nullable=True, index=True
     )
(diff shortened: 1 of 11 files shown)


hushline/model/field_definition.py
```diff
@@ -20,16 +20,20 @@ class FieldDefinition(Model):
     __tablename__ = "field_definitions"
     __table_args__ = (UniqueConstraint("username_id", "sort_order"),)
 
-    id: Mapped[int] = mapped_column(primary_key=True)
-    username_id: Mapped[int] = mapped_column(db.ForeignKey("usernames.id"), index=True)
+    id: Mapped[int] = mapped_column(primary_key=True, nullable=False)
```
brassy-endomorph, line 23: all the int/PK columns can have auto_increment added which may help silence some warnings? I think this is the preferred thing to do anyway for clarity.

hushline/model/field_definition.py
```diff
@@ -20,16 +20,20 @@ class FieldDefinition(Model):
...
     username: Mapped["Username"] = relationship(back_populates="message_fields")
-    label: Mapped[str] = mapped_column(db.String(255))
-    field_type: Mapped[FieldType] = mapped_column(SQLAlchemyEnum(FieldType))
-    required: Mapped[bool] = mapped_column()
-    enabled: Mapped[bool] = mapped_column()
-    encrypted: Mapped[bool] = mapped_column()
-    choices: Mapped[list[str]] = mapped_column(type_=JSONB)
-    sort_order: Mapped[int] = mapped_column()
+    label: Mapped[str] = mapped_column(db.String(255), nullable=False)
+    field_type: Mapped[FieldType] = mapped_column(
```
brassy-endomorph, line 29: This might get things in alignment, but it seems wrong to allow these to be nullable at all. I guess this is something to open a followup ticket for.

hushline/model/field_value.py
```diff
@@ -36,13 +36,17 @@ def add_padding(value: str, block_size: int = 512) -> str:
 class FieldValue(Model):
     __tablename__ = "field_values"
 
-    id: Mapped[int] = mapped_column(primary_key=True)
-    field_definition_id: Mapped[int] = mapped_column(db.ForeignKey("field_definitions.id"))
+    id: Mapped[int] = mapped_column(primary_key=True, nullable=False)
+    field_definition_id: Mapped[int] = mapped_column(
+        db.ForeignKey("field_definitions.id", ondelete="CASCADE"), nullable=True
```
brassy-endomorph, line 41: this cascade shouldn't be needed because nothing depends on a `FieldValue`, only the other way around

brassy-endomorph (commented):
nits mostly, i think

hushline/model/field_definition.py
micahflee: Yeah that does seem wrong. It would probably be fine to just change nullable to False and make a migration, but since I'm not entirely sure and don't want to break anything, let's do that in a future PR

hushline/model/field_value.py
micahflee: If I remove the cascade here, it tries to autogenerate this migration:

```py
def upgrade() -> None:
    # ### commands auto generated by Alembic - please adjust! ###
    with op.batch_alter_table('field_values', schema=None) as batch_op:
        batch_op.drop_constraint('fk_field_values_field_definition_id_field_definitions', type_='foreignkey')
        batch_op.create_foreign_key(batch_op.f('fk_field_values_field_definition_id_field_definitions'), 'field_definitions', ['field_definition_id'], ['id'])

    # ### end Alembic commands ###
```

So I think this is something else we should keep and fix in a future PR that includes an actual migration.

brassy-endomorph (approved):

PR #1086, opened 2026-01-09
Add Resend to Email button

glenn-sorrentino:
A few users have reported that a message has failed to send, and requested the ability to include a "Resend" button.

  This PR includes

  - a “Resend to Email” action on the message page that sends one email per field value, with a confirmation prompt to prevent accidental bulk sends
  - a hardened resend endpoint that validates CSRF and enforces ownership and notification settings before sending
  - new tests covering resend behavior, cross‑user access blocking, notifications disabled, and CSRF enforcement

## Details

When a message is originally sent, we concatenate the fields and encrypt on the front-end, making a message to Proton or Thunderbird easily decryptable. 

When we use the resend feature, messages are stored in fields in a PGP encrypted format. Field A is encrypted, field B, too. Because of this process, we cannot resend a message back in the same format as initially since we encrypt on the front end initially. 

The only way to deliver messages in a way that decrypts in clients where a private key is stored, we have to strip the label from the message, and send each field individually. This'll result in multiple emails, so a dialog is presented to confirm if they want this to happen. 

Edge cases of a client with 10+ fields may result in additional server load which we should monitor.


hushline/routes/message.py
```diff
@@ -79,6 +83,55 @@ def delete_message(public_id: str) -> Response:
 
         return redirect(url_for("inbox"))
 
+    @app.route("/message/<public_id>/resend", methods=["POST"])
+    @authentication_required
+    def resend_message(public_id: str) -> Response:
```
brassy-endomorph, line 88: So much of this seems to be duplicated logic from elsewhere in the app. The logic that create the email message from the tip components should exist in only once place so that the output is identical for both the original email being forwarded and the resend.

tests/test_resend_message.py
```diff
@@ -0,0 +1,189 @@
...
+    user.email_include_message_content = True
+    user.email_encrypt_entire_body = False
+    user.email = "test@example.com"
+    db.session.commit()
+
+    message = Message(username_id=user.primary_username.id)
+    db.session.add(message)
+    db.session.flush()
+
+    for field_def in user.primary_username.message_fields[:2]:
```
brassy-endomorph: Why are you are you truncating the list? At least leave a comment

tests/test_resend_message.py
```diff
@@ -0,0 +1,189 @@
...
+    db.session.commit()
+
+    sent = []
+
+    def fake_send_email(sent_user: User, body: str) -> None:
+        sent.append((sent_user.id, body))
+
+    monkeypatch.setattr("hushline.routes.message.do_send_email", fake_send_email)
+
+    csrf_token = _csrf_token_from_message_page(client, message.public_id)
```
brassy-endomorph: Is there a reason we extract the CSRF token here like this instead of just disabling it entirely for tests?

tests/test_resend_message.py
```diff
@@ -0,0 +1,189 @@
...
+    )
+
+    assert response.status_code == 200
+    assert "Message resent to your email inbox" in response.text
+    assert len(sent) == 2
+    assert all(sent_user_id == user.id for sent_user_id, _ in sent)
+
+
+@pytest.mark.usefixtures("_authenticated_user")
+def test_resend_message_blocks_other_users_messages(
```
brassy-endomorph, line 58: Can you leave a comment explaining the buisness requirement here? I can't figure out from the code or the test name what exactly is happening here. The only thing I can see different from the above test is that it's missing `    user.email_encrypt_entire_body = False` which doesn't make it clear what's being "blocked" or why that would be a requirement related to this feature.

tests/test_resend_message.py
```diff
@@ -0,0 +1,189 @@
...
+    db.session.commit()
+
+    sent = []
+
+    def fake_send_email(sent_user: User, body: str) -> None:
+        sent.append((sent_user.id, body))
+
+    monkeypatch.setattr("hushline.routes.message.do_send_email", fake_send_email)
+
+    prior_csrf_setting = app.config.get("WTF_CSRF_ENABLED")
```
brassy-endomorph: this should use `monkeypatch` or the `pytest-mock` fixture `mocker` to auto-cleanup at the end of a test

hushline/routes/message.py
glenn-sorrentino: The issue is that when we encrypt on the front end, we can concatenate and encrypt the entire body so clients like Proton can decrypt locally. Once we store the encrypted fields in the DB, we can't recreate the same layout because it double encrypts the fields, causing the decrypted contents to be cypiertext.

tests/test_resend_message.py
glenn-sorrentino: This was to harden the tests to make sure we're only sending messages to the intended user.

tests/test_resend_message.py
glenn-sorrentino: Updated the test file.

tests/test_resend_message.py
glenn-sorrentino: Updated the test file.

tests/test_resend_message.py
glenn-sorrentino: Updated the test file.

PR #1772, opened 2026-03-20
Replace gh project subcommands in issue runner

hushline-dev:
## Summary
- replace `gh project item-list` with GraphQL-based project item selection in the daily issue runner
- replace `gh project item-edit` with the `updateProjectV2ItemFieldValue` GraphQL mutation
- add focused runner tests for project item filtering and status updates

## Why
The current `gh` token now has Projects scope again, but `gh project ... --owner scidsg` is still failing with `unknown owner type`. This patch removes the runner's dependency on those flaky subcommands and uses the GraphQL API path that is already working.

## Verification
- `bash -n scripts/agent_daily_issue_runner.sh`
- shell-level verification that status updates call `updateProjectV2ItemFieldValue`
- shell-level verification that candidate selection only returns open issues in the target project status

## Note
I could not run the Python test file through the repo's normal `pytest`/`poetry` harness in the temp worktree because those tools are not installed in this environment.

### DIFF
--- a/scripts/agent_daily_issue_runner.sh
+++ b/scripts/agent_daily_issue_runner.sh
@@ -816,31 +816,215 @@ set_issue_project_status() {
     return 0
   fi
 
-  gh project item-edit \
-    --id "$item_id" \
-    --project-id "$project_id" \
-    --field-id "$field_id" \
-    --single-select-option-id "$option_id" >/dev/null
+  gh api graphql \
+    -f projectId="$project_id" \
+    -f itemId="$item_id" \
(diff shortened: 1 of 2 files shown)


hushline-dev:
Addressed in `db3ca27e`: `collect_issue_candidates_from_project` now paginates project items page-by-page and stops only after collecting up to `PROJECT_ITEM_LIMIT` matching open issues in the target status. I also added a regression test covering the multi-page case that previously would have been skipped after truncation.

PR #2193, opened 2026-06-11
Prevent alias directory location disclosure

glenn-sorrentino:
### Motivation
- The public directory builder previously copied account-level `User` location fields into every returned `Username` row, which allowed a public alias to leak the primary account's city/country/subdivision via `/directory/users.json`.
- This is a privacy-sensitive metadata disclosure affecting unauthenticated public directory rows and location filter metadata derived from visible `Username` objects.

### Description
- Stop exposing account geography for non-primary usernames by changing `_username_geography` to return an empty geography when `username.is_primary` is false in `hushline/routes/directory.py`.
- Add a regression test `test_directory_users_json_excludes_account_location_from_alias_rows` in `tests/test_directory.py` and import `Username` to assert that a public alias does not include account location when the primary username is hidden.
- Preserve existing behavior for primary usernames so normalized location and filter metadata continue to work for primary accounts.

### Testing
- `poetry run ruff format --check` and `poetry run ruff check` passed for the modified files and `poetry run mypy` reported no issues for the changed modules.
- Attempted `poetry run pytest tests/test_directory.py -q -k 'normalized_user_location or excludes_account_location_from_alias_rows'` but the run could not complete due to a test database host resolution error (`psycopg.OperationalError: Temporary failure in name resolution`) in this environment. 
- `make lint`, `make test`, and dependency audit targets were blocked in this environment because Docker is not available, so full CI-style checks must be run by maintainers before merge.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a2b356921ec8330b439d9c955ff0798)

### DIFF
--- a/hushline/routes/directory.py
+++ b/hushline/routes/directory.py
@@ -380,6 +380,9 @@ def _geography_fields(
 
 
 def _username_geography(username: Username) -> DirectoryListingGeography:
+    if not username.is_primary:
+        return build_directory_geography()
+
     user = username.user
     return build_directory_geography(
         city=getattr(user, "city", None),
--- a/tests/test_directory.py
+++ b/tests/test_directory.py
@@ -21,6 +21,7 @@ from hushline.model import (
     PublicRecordListing,
     SecureDropDirectoryListing,
     User,
+    Username,
     get_globaleaks_directory_listings,
     get_newsroom_directory_listings,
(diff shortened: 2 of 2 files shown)


glenn-sorrentino:
users manually add location data and opt-in to the directory. this isn’t an issue.

PR #2226, opened 2026-06-14
Add E2EE two-way conversations

glenn-sorrentino:
## Summary
- Add E2EE two-way conversations backed by chat keys, encrypted conversation copies, unread/presence state, and migration coverage.
- Update Inbox to show tips and conversations together, with Tips and Conversations filters in the Inbox nav.
- Place inbox row dates in a shared top-right meta position for both tips and conversations.
- Place the unread conversation dot inline before the `From:` label, so unread rows read visually as `• From: @username`.
- Add a message-id read cursor for conversations so unread state is robust when multiple messages share the same timestamp.
- Poll the open Inbox view for list and tab-count updates so incoming conversation activity can surface without a manual refresh.
- Keep background conversation-thread refreshes from marking incoming messages read, so the Inbox unread dot remains visible until the user intentionally opens the thread.
- Keep conversation presence heartbeats CSRF-protected while sourcing the token from the conversation root instead of depending only on the composer form.
- Refresh the conversation UI into a traditional chat thread with left/right bubbles, fixed header, pinned bottom composer, live in-place refresh for incoming messages, Enter-to-send, and mobile edge-to-edge container styling with the parent border removed at 100% width.
- Add a Back to Inbox link above the conversation display name in the fixed conversation header.
- Move the E2EE badge out of the header and into the bottom of the chat window, centered behind the message bubbles at 0.95x scale.
- Hide the E2EE badge until the browser chat key is unlocked; with JavaScript disabled, chat stays disabled and the UI states there is no server-side chat fallback.
- Make the conversation header identify the other participant instead of the logged-in recipient, avoiding the appearance that the user is chatting with themself.
- Make the conversation scroll surface span the browser width so the scrollbar sits at the viewport edge while header, message bubbles, and composer controls keep a centered readable inset.
- Create the chat key during successful login from a browser-generated, password-wrapped payload, including after successful 2FA, so the settings page should not keep showing a missing-key state after login.
- Unlock and restore the chat key from the authenticated browser session so conversations do not ask for a second password or show the dead-end `log out and log back in` lock state.
- Allow a logged-in user to start a chat with a recipient who has an automatic chat key even when that recipient has no PGP key.

## Why
Two-way communication needs to preserve Hush Line's E2EE expectations while feeling like a normal chat interface. This keeps message bodies client-encrypted for conversation participants and removes the extra in-conversation password prompt once the user is already authenticated and their chat key is unlocked for the session.

Recipients may not have uploaded a PGP key, but the chat key is provisioned automatically. Logged-in users should still be able to start an E2EE chat with those recipients, while anonymous tips and embedded forms continue to require the existing recipient encryption target.

### Protocols selected
- **Key agreement + encryption**: `ECDH-P256-AES-GCM`.
  - Per-message ephemeral ECDH key pair generated in browser for each encryption.
  - Shared secret derived via P-256 ECDH and used as AES-GCM key.
  - Random 12-byte IV and authenticated encryption for message confidentiality/integrity.
- **Integrity/authenticity**: ECDSA over P-256 with SHA-256 for v2 envelopes.
  - Public signing key is created with chat key generation and stored with the chat key record.
  - Message envelopes are signed in a canonical JSON payload before being sent.
- **Context binding**: v2 envelope includes immutable context fields (`purpose`, `conversation_id`, `sender_participant_id`, `recipient_participant_id`, key version/fingerprint).
  - Context is signed and used as AES-GCM AAD on decrypt/encrypt path.
  - Server enforces algorithm/version and verifies context binding before accepting payloads.
- **Server-side storage hardening**: chat private key bundles are wrapped client-side with **PBKDF2-SHA-256 + AES-256-GCM** before upload.
  - Chat key payloads include key derivation parameters, salt, and ciphertext in encrypted envelope form.
- **Backward compatibility**: legacy v1 envelopes continue to decrypt, while malformed v2 envelopes or invalid context are rejected.

### Why this stack was chosen
- **Fits Hush Line's deployment model**: all cryptography runs in browser, and servers only route encrypted blobs + validate metadata.
  - No server-side plaintext message channel, no added key-escrow component.
- **WebCrypto + Python ecosystem compatibility**: P-256 JWK and AES-GCM are widely supported in browser WebCrypto, reducing dependency risk and avoiding native crypto dependencies.
- **Good security-to-complexity balance**:
  - Per-message ephemeral ECDH gives message-level forward-secrecy characteristics.
  - AES-GCM provides confidentiality and integrity in one primitive.
  - ECDSA signatures and AAD significantly reduce message substitution/replay risks.
- **Operational resilience**: canonicalized/signed envelopes and versioned formats (`v`, `v2`) support safe protocol evolution in production without breaking existing data.
- **Clear user-key lifecycle**: rotation is explicit with versioned fingerprints and recipient warning UI when participants are on stale keys.

### Why not alternatives (for this phase)
- RSA-based transport was not selected due larger key/ciphertext overhead and poorer fit for per-message ephemeral patterns.
- Advanced ratcheting (Signal/X3DH-like) and PQ transitions were not adopted in this step to avoid introducing larger protocol complexity and deployment risk while product needs remain server-mediated inbox-style messaging.
- Non-P-256 curves were deprioritized to preserve JWK interoperability across current client stacks and reduce migration surface in a security-critical codebase.

### Security properties introduced in this PR
- Conversation posts require encrypted copies for every participant and reject plaintext-only payloads.
- Conversation chat has no server-side plaintext fallback: if JavaScript is disabled, composer remains disabled and the user sees a clear JS-required explanation.
- Login-time chat key provisioning stores public key plus password-encrypted private-key payload only; plaintext key material is never persisted server-side.
- Chat key unlock state is browser-session scoped (12-hour expiry) with cross-tab sharing; stale material is auto-expired and key-specific.
- CSRF handling for presence heartbeats now derives token from the conversation root with composer/global fallback.
- no-PGP chat-only profile starts are limited to authenticated non-self non-embedded submissions and require encrypted recipient copies.
- Anonymous/self/embedded/PGP-required flows still require recipient encryption target outside no-PGP chat-only path.

## Security
- Migrations keep chat/conversation tables free of plaintext body/private-key columns and enforce key constraints/cascades.
- `conversation_presence` CSRF checks are preserved while avoiding false 400s on valid conversation pages.
- Read-state model now supports message-id cursors to avoid unread collisions on same-timestamp events.
- Background polling never advances read cursor, while opening a conversation does.
- Chat badge visibility is tied to browser unlock state so UI does not claim encrypted state without keys.

## Validation
- `npx webpack --config webpack.config.js --env WEBPACK_WATCH=true` (passes with existing Sass legacy JS API deprecation warning)
- `docker compose run --rm app poetry run pytest tests/test_profile.py::test_logged_in_profile_submit_without_recipient_pgp_uses_chat_only_conversation tests/test_profile.py::test_logged_in_profile_submit_without_pgp_requires_valid_chat_payload tests/test_profile.py::test_logged_in_profile_submit_message_requires_recipient_encryption_target tests/test_profile.py::test_profile_pgp_required tests/test_frontend_compat.py::test_client_side_encryption_prepares_chat_conversation_copies -q` (`5 passed`)
- `docker compose run --rm app poetry run pytest tests/test_conversation.py::test_conversation_presence_accepts_rendered_csrf_token tests/test_conversation.py::test_conversation_presence_requires_csrf_when_enabled tests/test_frontend_compat.py::test_conversation_presence_uses_root_csrf_token -q` (`3 passed`)
- `docker compose run --rm app poetry run pytest tests/test_conversation.py::test_conversation_view_shows_locked_chat_key_state tests/test_frontend_compat.py::test_chat_key_lifecycle_restores_unlocked_key_for_authenticated_browser_session tests/test_frontend_compat.py::test_conversation_secure_badge_sits_behind_chat_messages -q` (`3 passed`)
- `make lint` (passes)
- `make test` (`2073 passed, 1 deselected, 1 xfailed, 11 warnings`)
- `make audit-python` (`No known vulnerabilities found`)
- `make audit-node-runtime` (`found 0 vulnerabilities`)

## Manual Testing
- Log out, log back in as a seeded recipient with no chat key, open Settings -> Encryption, and confirm the active chat-key version appears instead of the missing-key message.
- Repeat the login check with a 2FA-enabled account and confirm the key appears only after the OTP succeeds.
- After login, open or reload a conversation and confirm there is no `Chat key locked` panel, no `log out and log back in` instruction, and messages decrypt without a second password prompt.
- With JavaScript disabled, open a conversation and confirm the composer is disabled, the E2EE badge is not shown, and the page explains that E2EE chat requires JavaScript with no server-side chat fallback.
- Open Inbox and confirm tips and conversations appear in one chronological list.
- Confirm tip and conversation dates sit in the top-right corner of each inbox row.
- Confirm unread conversation rows show the dot directly before `From: @username`.
- With Inbox open in one account, send a conversation message from the other account and confirm the unread dot appears without manually refreshing.
- With a conversation open in one account, send a message from the other account, then return to Inbox and confirm the unread dot is still shown until the conversation is intentionally opened from Inbox.
- Use the Inbox nav filters for All, Tips, and Conversations and confirm each view shows the expected item types.
- Open a conversation and confirm the header shows the other participant's display name/username, not the logged-in account.
- Confirm the Back to Inbox link appears above the participant display name in the fixed conversation header.
- Open a conversation and confirm it renders as a chat-style thread with incoming messages on the left, recipient replies on the right, the header fixed under the global header/banner, and the one-line composer pinned at the bottom.
- Confirm the E2EE badge is centered near the bottom of the chat window, behind message bubbles, and only appears after the chat key is unlocked.
- Confirm the conversation presence request no longer logs a 400 immediately after loading a valid conversation page.
- Confirm the conversation scrollbar is on the browser edge, not immediately beside the message content, while the message bubbles remain inset and readable.
- At mobile width, confirm the conversation container spans the viewport and the parent container border is gone.
- Send a reply with Enter and confirm the new message appears at the latest position without staying below the fold.
- Open the same conversation in another tab while still logged in and confirm no separate chat password prompt is required after the chat key is browser-session-unlocked.
- As a logged-in sender, visit a recipient profile that has an automatic chat key but no PGP key and confirm the form is enabled, submission redirects into a conversation, and the conversation decrypts.
- As an anonymous sender or embedded form visitor, confirm a no-PGP recipient remains blocked from normal tip submission.

## Risks / Follow-ups
- Rotating a Hush Line chat key intentionally makes older conversations encrypted to previous keys unreadable.
- Browser-side decrypted chat key material is now retained across tabs/reloads for the authenticated browser session with a 12-hour expiry; this matches the no-second-password requirement but reinforces the importance of XSS/CSP controls.
- Existing SQLAlchemy fixture warnings remain in tests around transient objects during conversation setup and password-reset chat-key setup; they do not fail the suite.
- Fingerprints and rotated-key warnings improve manual awareness but are not a key-transparency log.
- Legacy chat keys without signing keys continue using v1 readable envelopes until users rotate/provision new signing-capable chat keys.


glenn-sorrentino:
Update after running `make test`: full suite completed at 100% collection/run and exited with infra failures after long run, not a code regression.

- Result: `1932 passed, 1 deselected, 1 xfailed, 8 warnings, 150 errors in 540.09s` (exit 1)
- Error pattern: Postgres became unavailable during run (`psycopg.errors.DiskFull` / `could not resize shared memory segment ... No space left on device`) and then switched into recovery state (`FATAL: database system is in recovery mode`, `not yet accepting connections`).
- Tests were otherwise executing and passed until DB recovery took effect.

Given this, the current branch changes remain as previously implemented. I can re-run `make test` after a fresh DB reset (`docker compose down -v --remove-orphans` + `docker compose up -d postgres blob-storage` + `docker compose run --rm dev_data`) and report a clean result once infra is stable.

glenn-sorrentino:
Update: pushed latest changes on branch `codex/fix-epic-2177-chat-qa`.

Latest commit: `396d2859` (signed).

Ran full suite from fresh DB reset:
- `docker compose down -v --remove-orphans`
- `docker compose up -d postgres blob-storage`
- `docker compose run --rm dev_data`
- `make test`

Result: `2082 passed, 1 deselected, 1 xfailed` in 415.87s (`0:06:55`), no failures.

Warnings observed are SQLAlchemy fixture warnings only (no test failures).

glenn-sorrentino:
Update: `lint` now passes after latest push.

Commit: `bca97d14` (signed)

I fixed three remaining lint issues from previous state:
- `ruff format` on `hushline/routes/message.py` and `tests/test_conversation.py`
- shortened long test function name in `tests/test_conversation.py` (E501 fix)
- ran prettier write on `hushline/static/js/directory_verified.js` and `hushline/static/js/settings-location.js`
- added `docs/screenshots/**/*.png` to `.prettierignore` to avoid container-specific binary read failures during prettier checks

`make lint` is now green.

